PR #65, opened 2023-05-06
Flexible spectrogram height

777arc:
fills the screen height, when you are scrolled all the way up at least


Nepomuceno (approved):
:shipit:

PR #95, opened 2023-05-12
Add pre-commit config + first pre-commit fix

hannahlspencer:


Nepomuceno:
I dont think pre-commit should be in the requirments txt of the detectors would make more sense to have the instructions on how to install it in the pre-commit becuase it si nto an actuall requiremwnr for the project it just helps you to spot mistakes earlier on in the process. THe guarantee it si actually done at the pull request time.

777arc:
> I dont think pre-commit should be in the requirments txt of the detectors would make more sense to have the instructions on how to install it in the pre-commit becuase it si nto an actuall requiremwnr for the project it just helps you to spot mistakes earlier on in the process. THe guarantee it si actually done at the pull request time.

agree, can remove pre-commit from requirements.txt

Nepomuceno (approved):
Great addition to the repo

PR #96, opened 2023-05-12
basic ci pipeline 

sebastus:
This basic github workflow runs on any "push" event in any branch.

It's based on Marc's original pipeline "main_iqengineweb.yml", but rather than deploying to the staging slot of the app service, it creates a container image and pushes it to ghcr.io container registry.

I'm not sure about the first step - where the app is built by npm. Do we plan to test that build, or test the containerized build, or are there some tests for each?

### DIFF
--- a/.github/workflows/ci-to-container.yml
+++ b/.github/workflows/ci-to-container.yml
@@ -56,5 +56,5 @@ jobs:
         with:
           push: true  
           tags: |
-            ghcr.io/${{ steps.set_variables.outputs.repository }}:${{ steps.set_variables.outputs.sha }}
+            ghcr.io/iqengine:v1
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-to-container.yml
```diff
@@ -0,0 +1,65 @@
+name: Build, test, containerize and store in registry Node.js app
+
+on: [push, workflow_dispatch]
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+
+    steps:
```
Jaydee123, line 10: I don't think we need this step as it's already covered in main_iqengineweb.yml

Jaydee123: In fact I think the containerisation is also covered in docker-release.yml

PR #97, opened 2023-05-13
Tailwind for spectrogram page and navbar

777arc:


client/src/App.jsx
```diff
@@ -31,57 +31,48 @@ export const App = () => {
...
-                <a
-                  className="nav-link"
-                  style={{ display: 'inline-flex', fontSize: '24px', paddingLeft: '0px', paddingRight: '20px' }}
-                  target="_blank"
-                  rel="noreferrer"
-                  href="https://discord.gg/k7C8kp3b76"
-                >
-                  <img src="/discord.svg" style={{ width: '22px' }} alt="Discord" />
-                  Discord
+          <div className="absolute right-0 py-2 mr-2">
```
Jaydee123, line 39: Is it worth putting these into their own 'Header.tsx' file?

777arc: yep good point

777arc: looks like we already have navbar and footer components but they are only used for the v2 version, pretty soon we'll merge the two versions into one and then we can reorganized this stuff

PR #101, opened 2023-05-15
Add vite instructions

hannahlspencer:


Nepomuceno:
Should't we add those libraries ad dev dependencies?

PR #102, opened 2023-05-15
add testing workflow for client

sebastus:
Added gh workflow that runs 'npm run test' in the client folder.  It's triggered by PR to main.


.github/workflows/run_tests.yml
```diff
@@ -0,0 +1,29 @@
+name: Run client tests
+
+on:
+  pull_request:
+    # only runs on PRs that target main
+    branches:
+      - main
```
Nepomuceno, line 7: We could add a path filter to just run on changes to /client
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

sebastus: added

PR #103, opened 2023-05-15
Simple pytest-flask test for api

PaulDFoster:
Add a simple pytest-flask test for the api

### DIFF
--- a/api/tests/conftest.py
+++ b/api/tests/conftest.py
@@ -1,0 +1,24 @@
+import pytest
+from flask import Flask
+
+def create_app(test_config=None):
+    app = Flask(__name__, static_folder='./build', static_url_path='/')
+
+    @app.route('/api/status')
+    def get_status():
+        return "OK"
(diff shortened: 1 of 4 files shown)


api/tests/conftest.py
```diff
@@ -0,0 +1,24 @@
+import pytest
+from flask import Flask
+
+def create_app(test_config=None):
+    app = Flask(__name__, static_folder='./build', static_url_path='/')
+
+    @app.route('/api/status')
```
Jaydee123, line 7: Don't you want to reference the original reference instead of copy the code here?

Nepomuceno (approved):
LGTM
Did you check if this did not broke the container?

Nepomuceno (approved):
LGTM
Did you check if this did not broke the container?

Nepomuceno (approved):
LGTM
Did you check if this did not broke the container?

PR #112, opened 2023-05-16
Paulfo/716 fix dockerfile

PaulDFoster:
Corrected the api app reference in the dockerfile
Change references the create_app() function

Tested with local docker build and run, checking localhost:3000/api/status and container log file. All ok.


sebastus (approved):
LGTM

PR #115, opened 2023-05-16
Fix lint errors in readme and alter workflow checks

Jaydee123:


.github/workflows/most_recent_pr_iqengineweb.yml
```diff
@@ -42,7 +42,7 @@ jobs:
           path: client/build/
 
   deploy-to-most-recent-pr-slot:
-    if: github.repository_owner == 'IQEngine'
+    if: github.event.pull_request.head.repo.full_name == 'IQEngine/IQEngine'
```
Nepomuceno, line 45: SHouldn't we move this to the top of the JOb there is no point on doing anything of this job if this is the case.

Nepomuceno (commented):
My understanding it is that this will stop any PR that does not come from the same repo that IQEngine is ( this menas any fork ) not to deploy to the latest pr slow if this is what we are thinking about doing perfect

.github/workflows/most_recent_pr_iqengineweb.yml
Jaydee123: Conditionals work at job level so have added to other job also:
https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution

Nepomuceno (approved):
:shipit:

PR #118, opened 2023-05-16
Add test

Jaydee123:


Shuave60:
I'm using Python 3.12 64 bit on a MacBookPro but I cannot run Python using neither Apple Terminal or Visual Studio IDE. Can someone please help to configure these CL so I can run LOGIC.py ?
Thanks

PR #129, opened 2023-05-22
First cut of Plugins API

samb-vt:
First cut of the Plugins API

### What does this PR do?

The plugins API is a unified interface tool for managing RF recording dataflow with the goal of standardizing inputs and outputs. The plugins API is planned to be converted to an OpenAPI spec.  Seeking feedback to enable further iteration!


777arc:
Gabe and I were chatting about what this API could be called.  From IQEngine's perspective it's like a plugin, but the hope would be for it to be used beyond IQEngine, and be a standard open interface for a variety of RF-style functions/processors like modems, detectors, etc.  One idea is to just call it the "IQEngine Open Interface" or similar, another idea is to come up with a new name/branding like RFprocessor API or something.  Definitely looking for ideas from folks, need to decide soon because the name ends up getting baked into code and is annoying to change later =)

Nepomuceno:
I would think that adding the sigmf file to the post to the function and making the return being a sigmf file too would make it standard and could make it more flexible and easier for anyone that is working with it because they acould just use any sigmf data reader.

777arc:
> I would think that adding the sigmf file to the post to the function and making the return being a sigmf file too would make it standard and could make it more flexible and easier for anyone that is working with it because they acould just use any sigmf data reader.

Because SigMF was designed to standardize metadata, and not as a REST API, it has tons of optional fields and very few required fields, so I don't think it makes sense to have this open interface directly using SigMF.  Plus there are mechanisms like Captures that will complicate the interface, and we're trying to keep it simple.  So the original idea was to reuse SigMF's annotations system as the way for detectors/classifiers to output their results.

777arc:
Another design decision- there will be plenty of RF Functions that take in IQ and output IQ, like a filter for example, but then there will also be stuff like modems that output bytes.  So on the output should we reuse the same field for both IQ and output bytes, and simply set data_type to char/uint8?

777arc:
filepath should probably be file_path for consistency

777arc:
Yet another thought - on one hand we could just not have an API and say that SigMF is the API, i.e. SigMF is the input and SigMF is the output.  But my opinion there is that SigMF is full of optional stuff and leaves a lot of room for flexibility, but with this type of API I think we want to be more prescriptive about the I/O, to give the interface more value and make it easier for authors to add it.  That might mean we dont support certain types of RF functions.  So there's a tradeoff between the two ends of the spectrum to be analyzed.

samb-vt:
My thoughts are that we probably want this API to have a more rigid definition than SigMF because having too much flexibility might make the API harder to adopt or less streamlined; i.e. we wouldn't want there to be too much ambiguity in the spec.

Here's a list of decisions we should make:

- Multiple RF inputs and possibly multiple outputs would be nice
   - The function would need to specify the number of inputs/outputs
   - Format could be a list for the input and output
   - Default would be 1 in, 1 out

- How should the byte-array output be dealt with?
   - We could reuse the same IQ format for byte arrays using uint8s
   - We could make a new format type specifically for byte arrays

- How do we send and receive large chunks of data through REST calls?

- Output should be able to have a different sample rate (default would be same as input)

777arc:
also different center freq for the output

777arc:
Alright I made those tweaks you pointed out @samb-vt 
- input IQ samples is now a list
- output now allows IQ samples and is also a list
- the initial get response includes the max input and output
- added sample_rate and center_freq to output

777arc:
Even though in the long term we wont want IQ to be passed within the REST call as a JSON array of ints/floats, maybe for this first cut of the API and server we can, to keep it simple, plus its the way the current detector works.  If someone doesn't want to pass a ton of data around within the call, they can just use the pointer to cloud storage.

And speaking of long term, right now these are synchronous calls, but we'll need an async way to do it for RF functions that take a while to run, e.g. you do the POST and then get returned the location where results (or progress %) can be found.  But once again, for now doing it synchronous is easier for the first cut of the server.

sebastus:
Has anyone given thought to plugins being cli apps that can use stdin/stdout + pointers to files as inputs and outputs? Could use json spec as input if it's a really complex use case. This would be very flexible - enough to be useful in both automated data ingestion pipelines and as "processor on request" from the UI. They would be easy to host in a container app.

777arc:
> 

You would still need a REST interface on top of the CLI though.  But not a bad idea, will have to think about it.

777arc:
@samb-vt  In order to support a variety of outputs in addition to IQ samples, we could leverage the existing MIME types, e.g. as a way for the RF Function to output an image, or sound file, or untyped-binary.  And we can create our own MIME type for iq, that uses SigMF's format, e.g. iq/ci16_le.   https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types 

We also thought it would be best for sample rate and center freq to be within the samples object, because on the input side there are situations where the center freqs dont match, like when you're doing bandwidth stitching.

777arc (approved):

PR #138, opened 2023-05-24
Add datasource api methods

Nepomuceno:
### What does this PR do?

This PR adds the stub for the calls to the backend api.

This implements it in a way that is transparent to the rest of the application if the data it is comming from each one of the iomplementations. This stub methods are here just to enable us to colaborate in the implementation of such methods.

### All Submissions:

### DIFF
--- a/client/src/api/datasource/ApiClient.ts
+++ b/client/src/api/datasource/ApiClient.ts
@@ -1,0 +1,25 @@
+import axios from 'axios';
+import { DataSourceClient } from './DataSourceClient';
+
+export class ApiClient implements DataSourceClient {
+  list() {
+    return axios.get('/api/datasources');
+  }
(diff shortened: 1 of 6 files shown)


client/src/api/datasource/ApiClient.ts
```diff
@@ -0,0 +1,25 @@
+import axios from 'axios';
+import { DataSourceClient } from './DataSourceClient';
+
+export class ApiClient implements DataSourceClient {
+  list() {
+    return axios.get('/api/datasource');
```
PaulDFoster, line 6: URL is /api/datasources

The plural is used across all the api

Nepomuceno: There will need to be changes on this for the actual implementaiton this are just stubs in place for the moment

PR #140, opened 2023-05-24
Build out the Server API

tobybrad:
### What does this PR do?

Converts API app to FastAPI
Adds basic CRUD-like operations for Datasources and Metadata

### All Submissions:

### Changes to Core Features:


api/handlers/datasources.py
```diff
@@ -0,0 +1,37 @@
...
+    return str(datasource_id)
+
+@router.get("/api/datasources")
+def get_datasources(db: object = Depends(database.database.db)) -> GetDatasourcesResponse:
+    datasources = db.datasources.find()
+    result = []
+    for datasource in datasources:
+        datasource['_id'] = str(datasource['_id']) 
+        result.append(datasource)
+    return {"datasources": result}
```
Nepomuceno, line 37: Wouldn't we just return one array here ?

api/handlers/metadata.py
```diff
@@ -0,0 +1,91 @@
...
+          'datasource_id': datasource_id,
+          'filepath': filepath,
+          'metadata': metadata
+      }
+      result = db.metadata.insert_one(initial_version)
+      db.versions.insert_one(initial_version)
+      return "Success"
+
+def get_latest_version(db, datasource_id, filepath):
+    cursor = db.versions.find({'datasource_id': datasource_id, 'filepath': filepath}).sort('version', -1).limit(1)
```
Nepomuceno, line 62: Why on this method not just use the db.metadata?

tobybrad: Syntax Error. Can you explain a bit more.

api/handlers/datasources.py
tobybrad: I think so. @PaulDFoster ?

.gitignore
```diff
@@ -1,5 +1,9 @@
 # See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
 
+.vscode
```
Nepomuceno, line 3: The idea of adding this files was to ake the config of the editor the same among developers to make easier for someone that is getting the project.

777arc:
If we remove .vscode/settings.json from the repo then stuff like prettier settings, like chars per line, wont be synced between devs and it will cause annoyances right?

api/handlers/metadata.py
```diff
@@ -0,0 +1,126 @@
...
+        "metadata": metadata,
+    }
+    db.metadata.insert_one(initial_version)
+    db.versions.insert_one(initial_version)
+    return "Success"
+
+
+def get_latest_version(db, datasource_id, filepath):
+    cursor = (
+        db.versions.find({"datasource_id": datasource_id, "filepath": filepath})
```
Nepomuceno, line 80: wouldn'r make sense here instead of getting the metadata from verions just getting it from metadata

db.metadata.find_one({"datasource_id": datasource_id, "filepath": filepath} )

api/handlers/datasources.py
```diff
@@ -0,0 +1,50 @@
...
+    db: object = Depends(database.database.db),
+):
+    """
+    Create a new datasource. The datasource will be henceforth identified by accountName_containerName which
+    must be unique or this function will return a 400.
+    """
+    if db.datasources.find_one(
+        {
+            "accountName": datasource.accountName,
+            "containerName": datasource.containerName,
```
Nepomuceno, line 32: Would not make senbse to add a field datasource_id that way makes a lot easier to search and link the metadta with the detasource later without the need to be converting and spliting things.

Nepomuceno:
> If we remove .vscode/settings.json from the repo then stuff like prettier settings, like chars per line, wont be synced between devs and it will cause annoyances right?

It would be caught up by the linter anywway but yes it would be caught way later in the process

api/handlers/datasources.py
PaulDFoster: Break down is required for the browse datasources page. It would make more sense to not have encoded meaning in the datasource_id have have the accountName/containerName in the URL specifically IMHO

api/handlers/metadata.py
PaulDFoster: As broad as it is long.

api/handlers/datasources.py
PaulDFoster: Yes that is correct. We can simplify to a single array

api/handlers/datasources.py
PaulDFoster: replaced datasource_id with explicit accountName and containerName

Jaydee123:
I think we can do the  following:
image
https://fastapi.tiangolo.com/advanced/additional-responses/

To display the expected responses via swagger?

api/handlers/datasources.py
```diff
@@ -0,0 +1,51 @@
...
+def create_datasource(
+    datasource: Datasource,
+    response: Response,
+    db: Database = Depends(database.database.db),
+):
+    """
+    Create a new datasource. The datasource will be henceforth identified by accountName/containerName which
+    must be unique or this function will return a 400.
+    """
+    if db.datasources.find_one(
```
Jaydee123, line 30: Do we have a unique index on account name and container name? Just thinking of a race condition where two requests are made at the same time, this would return null and two would be added

api/handlers/datasources.py
```diff
@@ -0,0 +1,51 @@
...
+        {
+            "accountName": datasource.accountName,
+            "containerName": datasource.containerName,
+        }
+    ):
+        response.status_code = 400
```
Jaydee123, line 36: Should this be 409?

api/handlers/metadata.py
```diff
@@ -0,0 +1,187 @@
...
+
+    # Check metadata doesn't already exist
+    if db.metadata.find_one(
+        {
+            "accountName": accountName,
+            "containerName": containerName,
+            "filepath": filepath,
+        }
+    ):
+        response.status_code = 400
```
Jaydee123, line 82: Should this be 409?

api/handlers/metadata.py
```diff
@@ -0,0 +1,187 @@
...
+def get_latest_version(db, accountName, containerName, filepath):
+    cursor = (
+        db.versions.find(
+            {
+                "accountName": accountName,
+                "containerName": containerName,
+                "filepath": filepath,
+            }
+        )
+        .sort("version", -1)
```
Jaydee123, line 115: How are the versions ordered? is it just numerical?

api/handlers/metadata.py
```diff
@@ -0,0 +1,187 @@
...
+        .limit(1)
+    )
+    result = list(cursor)
+    if not result:
+        return None
+    else:
```
Jaydee123, line 121: Is the else redundant here

api/handlers/metadata.py
```diff
@@ -0,0 +1,187 @@
...
+        {
+            "accountName": accountName,
+            "containerName": containerName,
+            "filepath": filepath,
+        }
+    )
+    if exists is None:
+        response.status_code = 404
+        return {"error": "Item not found"}
+    else:
```
Jaydee123, line 147: Same here? Not a massive fan of writing else if we return things from the previous if, I think it makes code less readable, however appreciate that's personal preference so if you disagree feel free to resolve this comment.

api/handlers/metadata.py
```diff
@@ -0,0 +1,187 @@
...
+        latest_version = get_latest_version(db, accountName, containerName, filepath)
+        if latest_version is None:
+            response.status_code = 500
+            return {"error": "No version found"}
+        with db.client.start_session() as session:
+            session.start_transaction()
+            try:
+                # This is going to be a race condition
+                version_number = latest_version["version_number"] + 1
```
Jaydee123, line 156: I assume this is a unique index? Just thinking what happens when someone tries to update at the same time, they would both end up with the same version number if not right?

api/handlers/metadata.py
```diff
@@ -0,0 +1,187 @@
...
+                current_version = db.metadata.find_one(
+                    {
+                        "accountName": accountName,
+                        "containerName": containerName,
+                        "filepath": filepath,
+                    }
+                )
+                if current_version is None:
```
Jaydee123, line 164: What does this do as apposed to 'get_latest_version' ?

api/handlers/metadata.py
Jaydee123: Also, don't we do the same call further up?
image

api/handlers/metadata.py
```diff
@@ -0,0 +1,187 @@
...
+                    response.status_code = 404
+                    return {"error": "Item not found"}
```
Jaydee123, line 166: Think this check is redundant based on other comment

api/handlers/datasources.py
Nepomuceno: We currently do not have given the current level of usage of the app I think this would not be a huge risk thart we are into but definetly something to keep in mind for the future. 
Curreently we do not have any databse model not as part of the application anyway.

api/handlers/metadata.py
Nepomuceno: I am writing this because the linter did nto agree with this being redundant.

api/handlers/metadata.py
Nepomuceno: Yeah this needs to be looked atr will chek this.

api/handlers/metadata.py
Nepomuceno: I think get_latest_version it is not needed now to be honest.

api/handlers/metadata.py
Nepomuceno: Yeah alsothought that the linter did not agree.

api/tests/test_api.py
```diff
@@ -0,0 +1,170 @@
...
+    client.post("/api/datasources", json=test_datasource).json()
+    response = client.post(
+        f'/api/datasources/{test_datasource["accountName"]}/{test_datasource["containerName"]}/file_path/meta',
+        json={"test": "string"},
+    )
+    response = client.get(
+        f'/api/datasources/{test_datasource["accountName"]}/{test_datasource["containerName"]}/file_path/meta'
+    )
+    assert response.status_code == 200
+    assert response.json()["version_number"] == 0
```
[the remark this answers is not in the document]
Nepomuceno: Wrote another test that checks the increment of the version number.

PaulDFoster (approved):
LGTM

Jaydee123 (approved):

Nepomuceno (approved):

PR #146, opened 2023-05-27
write load metadata script

sebastus:
### What does this PR do?

It adds a python script tool that gets metadata files from an azure blob container or ADLS file system and inserts them into the database using the api. This is response to task 747 in AzDO.

### All Submissions:

### Changes to Core Features:


tools/metadata_loader/main.py
```diff
@@ -0,0 +1,171 @@
...
+
+def call_create_meta_api(url, payload):
+    return requests.post(url, json=payload)
+
+def create_meta(accountName: str, containerName: str, filepath: str, document: str):
+
+    config = get_config()
+
+    #quoted_filepath = quote(filepath, safe='')
+    quoted_filepath = filepath.replace("/", "(slash)")
```
Nepomuceno, line 88: We will be using url encoding instead of just (slasj) to pass filenames there are plenty of other charss that are not url safe but that are encoded auttomatically by the url enconding function.

tools/metadata_loader/main.py
```diff
@@ -0,0 +1,171 @@
...
+    return overall_response
+
+
+def start():
+# commands
+# python main.py ...
+#    datasource add -name -accountName -containerName -description
+#    datasource list
+#    metadata list
+#    metadata addfolder -accountName -containerName -document
```
Nepomuceno, line 137: There is a convention in python for documentation strings called doc strings that would auytomatically add this to the method __doc__ that can be usefull for things like this.
https://peps.python.org/pep-0257/

tools/metadata_loader/main.py
```diff
@@ -0,0 +1,171 @@
...
+    for blob in blob_list:
+        
+        # print(blob.name)
+        # if blob.name == "/dir1/dir2/abc.sigmf-meta"
+        # then basename = abc.sigmf-meta, dirname = /dir1/dir2
+
+        basename = os.path.basename(blob.name)
+        parts = basename.split('.')
+        if len(parts)<2 or parts[1] != 'sigmf-meta':
+            continue
```
Nepomuceno, line 115: I imagine thiis is assuming that I have  blob storage that has only sigmf data because in the case of some other data sources youcould have also files with the same name but not sigmf extensions like .json or .csv I imagin there is the assumpotion that thsi will not be supported.

Nepomuceno (commented):
I would remove commented out code before finishing the pr but good tool
Did not run locally but looks good :) 
We could make sure that the api does get merged before becuase there would be a dependency of it for this to work.

tools/metadata_loader/main.py
sebastus: I tried quoting the string, but the api converted the %2F back to /. So in the interim, I changed it to (slash).

tools/metadata_loader/main.py
sebastus: Thanks for the reminder. done.

tools/metadata_loader/main.py
sebastus: I can add other file formats if samples are provided. But yes, at the moment I anticipate only the sigmf-meta file extension.

sebastus:
@Nepomuceno Maybe you noticed that the linter failed? The message said there is no docker folder. Which of course there isn't.

Nepomuceno:
The Linter does say that 3 of your files in python wouldbe reformated if you run the inter so might be a ggood idea to run it locally enabling the reformating.

tools/metadata_loader/main.py
```diff
@@ -0,0 +1,186 @@
...
+
+    return resp.text
+
+
+def call_get_all_metadata_api(url: str):
+    return requests.get(url, timeout=5)
+
+
+def get_all_meta(args):
+
```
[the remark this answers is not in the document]
sebastus: I use it to prove that the add folder routine worked.

.vscode/launch.json
```diff
@@ -1,9 +1,60 @@
...
   "configurations": [
+    {
+      "name": "Python: test metadata loader",
+      "type": "python",
+      "request": "launch",
+      "module": "pytest",
+      "cwd": "${workspaceFolder}/tools/metadata_loader"
+    },
+    {
+      "name": "datasource list",
```
Nepomuceno, line 12: Since we have just one launch.json would be good to identify these actions as being form the metadata loader something liek adding [LOADER] to the name.

tools/metadata_loader/main.py
```diff
@@ -0,0 +1,185 @@
...
+import requests
+from azure.storage.blob import (  # pyright: ignore[reportMissingImports]
+    BlobServiceClient,
+)
+from dotenv import load_dotenv
+
+
+def get_config():
+    load_dotenv()
```
Nepomuceno, line 14: Should we not put load dotenv to the begining of the file like line 11 ? doesnt need to be inside the function.

tools/metadata_loader/main.py
```diff
@@ -0,0 +1,185 @@
...
+    return {
+        "API_URL_BASE": os.environ.get("API_URL_BASE"),
```
Nepomuceno, line 16: I normally rather use os.getenv in case you do not wwant to test if the env exist.

tools/metadata_loader/main.py
```diff
@@ -0,0 +1,185 @@
...
+        "BLOB_STORAGE_ACCOUNT_URL": os.environ.get("BLOB_STORAGE_ACCOUNT_URL"),
+        "BLOB_STORAGE_SAS_KEY": os.environ.get("BLOB_STORAGE_SAS_KEY"),
+    }
+
+
+def call_get_datasources_api(url):
```
Nepomuceno, line 22: I dont get the why of these call methods that are jsut one line why not call requests directly in the function?

tools/metadata_loader/main.py
```diff
@@ -0,0 +1,185 @@
...
+    container_client = blob_service_client.get_container_client(
+        container=args.containerName
+    )
+
+    blob_list = container_client.list_blobs()
+
+    overall_response = True
+    for blob in blob_list:
+        basename = os.path.basename(blob.name)
+        parts = basename.split(".")
```
Nepomuceno, line 120: That would probably fail for `my.amazing.file.sigmf-meta`is this something we want to cover?

tools/metadata_loader/main.py
```diff
@@ -0,0 +1,185 @@
...
+            continue
+
+        blob_client = container_client.get_blob_client(blob=blob.name)
+        downloader = blob_client.download_blob(max_concurrency=1, encoding="UTF-8")
+        blob_text = downloader.readall()
+
+        dirname = os.path.dirname(blob.name)
+        filepath = f"{dirname}/{parts[0]}"
+
+        resp = create_meta(args.accountName, args.containerName, filepath, blob_text)
```
Nepomuceno, line 131: Are we going to creeate the metadata file even if there is no sigmf-data record in storage account?

tools/metadata_loader/main.py
```diff
@@ -0,0 +1,185 @@
...
+
+        overall_response = overall_response and resp == "Success"
```
Nepomuceno, line 133: why not only `overall_response = resp == "Success"`

.vscode/launch.json
sebastus: ok

tools/metadata_loader/main.py
sebastus: makes them easier to mock. If you have a better way, I'm all ears.

tools/metadata_loader/main.py
sebastus: ok

tools/metadata_loader/main.py
sebastus: ok

tools/metadata_loader/main.py
sebastus: I don't know what you mean. The program copies metadata from storage to the db. So if it's not in storage, it doesn't get to the db.

tools/metadata_loader/main.py
sebastus: I don't understand that code construct in python. what does it mean?

tools/metadata_loader/main.py
sebastus: fixed.

tools/metadata_loader/main.py
sebastus: that goes to the purpose of the loader. would you only run it after you have placed the IQ data where it belongs?

Nepomuceno (approved):
LGTM

PR #171, opened 2023-06-09
1001 add tests to pr

sebastus:
### What does this PR do?
This PR adds a github workflow called "Run Playwright frontend integration tests". The workflow is triggered on completion of the existing "Deploy to PR" workflow. It can also be triggered manually.

### All Submissions:

### Changes to Core Features:

### DIFF
--- a/playwright-pr.config.ts
+++ b/playwright-pr.config.ts
@@ -18,7 +18,7 @@ export default defineConfig({
   /* Retry on CI only */
   retries: process.env.CI ? 2 : 0,
   /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 3 : undefined,
+  workers: process.env.CI ? 9 : undefined,
   /* Timeout 90 seconds per test */
   timeout: 90000,
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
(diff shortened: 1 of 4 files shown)


Nepomuceno (approved):
Just more as an info this will not execute on any pr that it is done from a frok only on PRs that are done from branches insde the main repo that do target main ( because the deploy to pr does not run on forks ) 

Also we could mauybe remove the code that it comented

playwright-pr.config.ts
```diff
@@ -0,0 +1,79 @@
...
+      name: 'firefox',
+      use: { ...devices['Desktop Firefox'] },
+    },
+
+    {
+      name: 'webkit',
+      use: { ...devices['Desktop Safari'] },
+    },
+
+    /* Test against mobile viewports. */
```
Jaydee123, line 52: can we delete

playwright-pr.config.ts
```diff
@@ -0,0 +1,79 @@
...
+    //   use: { ...devices['Pixel 5'] },
+    // },
+    // {
+    //   name: 'Mobile Safari',
+    //   use: { ...devices['iPhone 12'] },
+    // },
+
+    /* Test against branded browsers. */
+    // {
+    //   name: 'Microsoft Edge',
```
Jaydee123, line 64: do we not want to test against edge?

sebastus: the chromium test targets edge. unless you think we need to target the name brand browser?

playwright-pr.config.ts
sebastus: done

PR #210, opened 2023-06-19
Min max to dB and simplified dualslider

777arc:


Nepomuceno (approved):
WOW just love this ❤️ reducingg a lot of the most complicated part of the code for me :)

777arc:
@Jaydee123 what do you think about those changes to your dualslider, having it no longer have its own state?  Seems to work fine, at least with this use-case.

Jaydee123:
> @Jaydee123 what do you think about those changes to your dualslider, having it no longer have its own state? Seems to work fine, at least with this use-case.

My only reservations is that a 'Dual Slider' component is meant to be generic, by referring to the 'Magnitude' we're tying it to that particular scenario, can we name it to 'updateMin' inside the class so it has more of a generic name?

client/src/Components/DualRangeSlider/DualRangeSlider.tsx
```diff
@@ -1,65 +1,64 @@
-import React, { useCallback, useEffect, useState, useRef } from 'react';
+import React, { useCallback, useEffect, useRef } from 'react';
 
-const DualRangeSlider = ({ min, minValue, max, maxValue, onChangeMin, onChangeMax }) => {
-  const [minVal, setMinVal] = useState(minValue);
-  const [maxVal, setMaxVal] = useState(maxValue);
+const DualRangeSlider = ({ min, minValue, max, maxValue, updateMagnitudeMin, updateMagnitudeMax }) => {
```
Nepomuceno, line 3:
```suggestion
const DualRangeSlider = ({ min, minValue, max, setValue, setMin, setMax }) => {
```
change this name would make the componetn a bit more generic.

client/src/Components/DualRangeSlider/DualRangeSlider.tsx
```diff
@@ -1,65 +1,64 @@
...
-  const [maxVal, setMaxVal] = useState(maxValue);
+const DualRangeSlider = ({ min, minValue, max, setValue, setMin, setMax }) => {
   const range = useRef(null);
-
   const getPercent = useCallback((value) => Math.round(((value - min) / (max - min)) * 100), [min, max]);
 
-  useEffect(() => {
-    const minPercent = getPercent(minVal);
-    const maxPercent = getPercent(maxVal);
+  // if some other component changes the max/min, update the slider visually
```
Jaydee123, line 7: can we remove

client/src/Components/DualRangeSlider/DualRangeSlider.tsx
```diff
@@ -1,65 +1,64 @@
...
+    const minPercent = getPercent(minValue);
+    const maxPercent = getPercent(maxValue);
     if (range.current) {
       range.current.style.left = `${minPercent}%`;
       range.current.style.width = `${maxPercent - minPercent}%`;
     }
-
-    onChangeMin(minVal);
-  }, [minVal, getPercent]);
+    updateMagnitudeMin(minValue);
```
Jaydee123, line 20: needs changing to setMin

client/src/Components/DualRangeSlider/DualRangeSlider.tsx
```diff
@@ -1,65 +1,64 @@
...
-
+    const minPercent = getPercent(minValue);
+    const maxPercent = getPercent(maxValue);
     if (range.current) {
       range.current.style.width = `${maxPercent - minPercent}%`;
     }
-
-    onChangeMax(maxVal);
-  }, [maxVal, getPercent]);
+    updateMagnitudeMax(maxValue);
```
Jaydee123, line 29: needs chaning to setMax

client/src/Components/DualRangeSlider/DualRangeSlider.tsx
```diff
@@ -1,65 +1,64 @@
...
         type="range"
         min={min}
         max={max}
-        value={minVal}
+        value={minValue}
         onChange={(event) => {
-          const value = Math.min(Number(event.target.value), maxVal - 1);
-          setMinVal(value);
+          const value = Math.min(Number(event.target.value), maxValue - 1);
+          updateMagnitudeMin(value);
```
Jaydee123, line 41: same here

client/src/Components/DualRangeSlider/DualRangeSlider.tsx
```diff
@@ -1,65 +1,64 @@
...
         type="range"
         min={min}
         max={max}
-        value={maxVal}
+        value={maxValue}
         onChange={(event) => {
-          const value = Math.max(Number(event.target.value), minVal + 1);
-          setMaxVal(value);
+          const value = Math.max(Number(event.target.value), minValue + 1);
+          updateMagnitudeMax(value);
```
Jaydee123, line 52: same here

PR #219, opened 2023-06-19
Dockerfile for plugins & repobrowser design tweaks

777arc:


Nepomuceno (approved):
Yeah the linter problems are just from moving the files.

PR #220, opened 2023-06-20
Misc cleanups

777arc:
- remove autoscale button
- fixed plugins spectrogram
- scale freq plot using fftsize
- added i8 and u8 datatypes
- added code mirror to metaraw
- added json syntax coloring to codemirror
- make an svg out of the local blob tile pic
- added back in jamies navbar when the window is less than 700px


777arc:
@Jaydee123  @Nepomuceno  I think this is an example of where unit testing stuff can really slow things down, I swapped from a textbox to codemirror which was a 5 min change and only a few lines of code to get it looking really nice, but its causing the MetaRaw.test.tsx unit test to fail and after spending like 10m trying to debug I gave up, there's some warning about running multiple code mirror instances at once...

Nepomuceno (approved):
Aparently React-codeMirror it is not compatible with vitest https://github.com/uiwjs/react-codemirror/issues/506

PR #235, opened 2023-06-21
disable trufflehog

777arc:
the way its yelling at the docker-compose line (shown below) makes me think trufflehog is more trouble than its worth, and there's no way to ignore files/lines with it.  so disabling it, unless someone objects.  we still have codeQL.

```
METADATA_DB_CONNECTION_STRING: mongodb://${MONGO_USERNAME}:${MONGO_PASSWORD}@mongo:27017
```


Nepomuceno:
I think that to disable trufflehog we could jsut delete the file.

Nepomuceno (approved):
Codeql takes care of this

PR #241, opened 2023-06-22
Colormap is now an option!

777arc:
the viridis colormap looks so sexy with our UI.  Note the colormap chooser on the left.

image


Nepomuceno (approved):
LGTM

PR #261, opened 2023-06-26
Refactor feature flags

hannahlspencer:
### What does this PR do?
Refactors feature flags so we have the option to hide components if we want to.

### All Submissions:

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -94,7 +94,7 @@ export const App = () => {
           </div>
           <Link to="/" onClick={() => {}}>
             <div id="IQEngineLogo" className="absolute mt-4 pt-2 top-0 left-1/2 transform -translate-x-1/2 flex">
-              <Feature flag="displayInternalBranding">
+              <Feature flag="hideInternalBranding" invert="true">
                 <img
                   src="/internalbrandingeg.jpg"
                   alt="Internal branding logo"
@@ -121,7 +121,7 @@ export const App = () => {
                   <div className="text-lg">Plugins</div>
                 </Link>
               </li>
-              <Feature flag="useIQEngineOutReach">
+              <Feature flag="useIQEngineOutReach" invert="false">
                 <li className="hidden md:block">
                   <a href="https://discord.gg/k7C8kp3b76" target="_blank" rel="noreferrer" className="text-lg">
                     <div className="flex">
@@ -149,7 +149,7 @@ export const App = () => {
                   </svg>
                 </a>
               </Feature>
-              <Feature flag="displayIQEngineGitHub">
+              <Feature flag="displayIQEngineGitHub" invert="false">
                 <li className="hidden md:block">
                   <a href="https://github.com/iqengine/iqengine" target="_blank" rel="noreferrer" className="text-lg ">
                     <div className="flex">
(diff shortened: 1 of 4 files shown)


777arc:
is there a way to make it so we dont need to specify invert, ie defaults to false

hannahlspencer:
> is there a way to make it so we dont need to specify invert, ie defaults to false

good point! Have added that in now

client/src/Components/Feature/Feature.tsx
```diff
@@ -3,13 +3,23 @@ import { useFeatureFlags } from '@/Components/FeatureFlagsContext/FeatureFlagsCo
 
 interface FeatureProps {
   flag: string;
+  invert: boolean | string;
   children: ReactNode;
 }
 
-const Feature = ({ flag, children }: FeatureProps) => {
+const Feature = ({ flag, invert, children }: FeatureProps) => {
```
Jaydee123, line 10: You should be able to pass in invert = false as a default here

client/src/Components/Feature/Feature.tsx
```diff
@@ -3,13 +3,23 @@ import { useFeatureFlags } from '@/Components/FeatureFlagsContext/FeatureFlagsCo
...
   const { getFeatureFlag } = useFeatureFlags();
+  if (!invert) {
```
Jaydee123, line 12: If you add above then this will be unneeded

client/src/Components/Feature/Feature.tsx
```diff
@@ -3,13 +3,23 @@ import { useFeatureFlags } from '@/Components/FeatureFlagsContext/FeatureFlagsCo
 
 interface FeatureProps {
   flag: string;
+  invert: boolean | string;
```
Jaydee123, line 6: Why not just make a boolean?

Jaydee123:
This provides a nice example of how we could default the values using boolean instead of string 😊
https://codesandbox.io/embed/elated-davinci-izut1?fontsize=14&hidenavigation=1&theme=dark&codemirror=1

client/src/App.tsx
```diff
@@ -94,7 +94,7 @@ export const App = () => {
           </div>
           <Link to="/" onClick={() => {}}>
             <div id="IQEngineLogo" className="absolute mt-4 pt-2 top-0 left-1/2 transform -translate-x-1/2 flex">
-              <Feature flag="displayInternalBranding">
+              <Feature flag="hideInternalBranding" invert="true">
```
Nepomuceno, line 97: I just fid this all too confusing to be honest reading this line I have no idea what will happen if I put hideInternalBranding as true will this hide the internal brand or display it because the invert is true ? I think would make more sense to add something like default and the default value for default would be true but you could set here a different value for default.

THinking about it would be much better to have a config file like the constants that we do have that would have the feature flags that are supported with the default value for them and them we would use that to set defaults instead of setting them here. the good thing of that approach it si that we aocul also use that to validate the allowed string values for feature flags to avoid misspelling and things like that.

client/src/Components/Feature/Feature.tsx
hannahlspencer: Get an error when trying to add it in the  <Feature flag="useIQEngineOutReach" invert="false"> without quotes, but with quotes it gets parsed as a string.

client/src/Components/Feature/Feature.tsx
hannahlspencer: Awesome, thanks for pointing that out!

client/src/Components/Feature/Feature.tsx
hannahlspencer: But with your link below I see how to do it, cheers!

hannahlspencer:
Closing as this has been re-refactored in https://github.com/IQEngine/IQEngine/pull/270

PR #268, opened 2023-06-27
switch to mongo for tests

sebastus:
### What does this PR do?
Implements a mongodb container as the database dependency for tests.

### All Submissions:

### Changes to Core Features:


e2e/teardown.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+docker kill iqengine
+docker kill iqenginedb
+docker network rm iqenginenet
+docker container prune --force
```
Nepomuceno, line 6: This command might endup removing a lor more than iqengine itself. would be easier to do something like docker compose up and down ?

e2e/webServer.sh
```diff
@@ -1,7 +1,38 @@
...
+  -e MONGO_INITDB_ROOT_USERNAME="$MONGO_INITDB_ROOT_USERNAME" \
+  -e MONGO_INITDB_ROOT_PASSWORD="$MONGO_INITDB_ROOT_PASSWORD" \
+  --name iqenginedb \
+  mongo:latest
+docker run --network iqenginenet -p 3000:3000 \
+  -e IN_MEMORY_DB=0 \
+  -e VITE_FEATURE_FLAGS='{"useAPIDatasources": true}' \
+  -e IQENGINE_METADATA_DB_CONNECTION_STRING="$conn" \
+  --name iqengine \
+  iqengine:latest
```
Nepomuceno, line 36: This would probably be a docker compose maybe use the same docker compose that spin up the plugins.

e2e/teardown.sh
sebastus: Switched to docker compose. I'm curious - how exactly does this remove more than intended?

e2e/webServer.sh
sebastus: can't use the same compose yml file. Need one that uses newly built image rather than the one on ghcr.io

e2e/teardown.sh
Nepomuceno: Container prune will remove not only the containers of IQEngine in  case you had other containers running

Nepomuceno: Since you are now using docker compose up wouldn't make sense to do a docker compose down in the teardown?

docker-compose-e2e.yml
```diff
@@ -0,0 +1,22 @@
...
+      - mongo
+    ports:
+      - 3000:3000
+  mongo:
+    image: mongo:4.2
+    restart: always
+    ports:
+      - 27017:27017
+    environment:
+      MONGO_INITDB_ROOT_USERNAME: ${MONGO_USERNAME}
```
hannahlspencer, line 21: Where do these variables come from?

hannahlspencer (approved):
Approving but had a question for my own knowledge

docker-compose-e2e.yml
Nepomuceno, line 1: Shouldn't this also contain the plugin image?

e2e/teardown.sh
sebastus: yes

docker-compose-e2e.yml
sebastus: they are mentioned in the env.sample, must be set by user in the environment

docker-compose-e2e.yml
sebastus: added

PR #283, opened 2023-06-29
create trivial plugin

sebastus:
### What does this PR do?
Adds a trivial plugin that enables testing of plugin output other than IQ data. It outputs a wav file.
Also fixed a bug with spelling of an env var in the plugins api.
Made handling of plugin output dependent on the mime type returned. If not 32bit IQ, it stores the file on disk. At present, only audio/wav is handled.
The new plugin has some notes about how to write a plugin.

### All Submissions:

### Changes to Core Features:


client/src/pages/spectrogram/components/PluginsPane.tsx
```diff
@@ -23,6 +24,9 @@ export interface PluginsPaneProps {
   setMeta: (meta: SigMFMetadata) => void;
 }
 
+const IQ_SIGMF_32 = 'iq/cf32_le';
```
Nepomuceno, line 27: Wouldn't be better to move this to an enum  like we did with FeatureFlag?

client/src/pages/spectrogram/components/PluginsPane.tsx
```diff
@@ -158,7 +162,33 @@ export const PluginsPane = ({ cursorsEnabled, handleProcessTime, meta, setMeta }
           });
 
           setModalOpen(true);
+        } else {
+          // Data file
+          const samples_base64 = data.data_output[0]['samples'];
+          const samples = Buffer.from(samples_base64, 'base64');
+
+          const a = document.createElement('a');
+          document.body.appendChild(a);
```
Nepomuceno, line 171: Do you need to append the element to the body ?

plugins/src/wav_out/wav_out.py
```diff
@@ -0,0 +1,42 @@
...
+        file_path = path.dirname(path.realpath(__file__))
+        with open(f"{file_path}/Windows Foreground.wav", mode="rb") as wavfile:
+            wav_file_content = wavfile.read()
+
+        samples_obj = {
+            "samples": b64encode(wav_file_content),
+            "sample_rate": self.sample_rate,
+            "center_freq": self.center_freq,
+            "data_type": "audio/wav",
+        }
```
Nepomuceno, line 41: Not that we need to do this for this request but looking at the api design sounds strange to have things like center frequency for other types of fgfiles liek that would that be for a text file for example.

sebastus: I agree, but the plugins_api requires those properties. We can change this in a future task.

client/src/pages/spectrogram/components/PluginsPane.tsx
sebastus: I don't know - I copied this code from the siggen page. It works.

client/src/pages/spectrogram/components/PluginsPane.tsx
```diff
@@ -94,7 +98,7 @@ export const PluginsPane = ({ cursorsEnabled, handleProcessTime, meta, setMeta }
           toast.error(`Plugin failed to run: ${data.status}`);
           return;
         }
-        if (data.data_output && data.data_output.length > 0) {
+        if (data.data_output[0]['data_type'] == IQ_SIGMF_32 && data.data_output && data.data_output.length > 0) {
```
Jaydee123, line 101: would it be better to do a find on data_type here which would return the first element with that data type?
const dataType = data.data_output.find(obj => obj['data_type'] === 'IQ_SIGMF_32');

client/src/pages/spectrogram/components/PluginsPane.tsx
```diff
@@ -158,7 +162,33 @@ export const PluginsPane = ({ cursorsEnabled, handleProcessTime, meta, setMeta }
...
           setModalOpen(true);
+        } else {
+          // Data file
+          const samples_base64 = data.data_output[0]['samples'];
+          const samples = Buffer.from(samples_base64, 'base64');
+
+          const a = document.createElement('a');
+          document.body.appendChild(a);
+          a.style = 'display: none';
+          var blob = new Blob([samples], { type: data.data_output[0]['data_type'] }),
```
Jaydee123, line 173: You would usually use const or let in react:
https://sentry.io/answers/difference-between-let-and-var-in-javascript/#:~:text=The%20difference%20between%20let%20and,in%20which%20they're%20declared.

client/src/pages/spectrogram/components/PluginsPane.tsx
sebastus: Done

client/src/pages/spectrogram/components/PluginsPane.tsx
sebastus: Ordinarily I would agree with you. But there's a comment in the code saying that 99% of plugins have only one return object, so handling of higher elements is disregarded throughout.

client/src/pages/spectrogram/components/PluginsPane.tsx
sebastus: done.

client/src/pages/spectrogram/components/PluginsPane.tsx
```diff
@@ -158,7 +168,33 @@ export const PluginsPane = ({ cursorsEnabled, handleProcessTime, meta, setMeta }
...
+          const samples = Buffer.from(samples_base64, 'base64');
+
+          const a = document.createElement('a');
+          document.body.appendChild(a);
+          a.style = 'display: none';
+          const blob = new Blob([samples], { type: data.data_output[0]['data_type'] }),
+            url = window.URL.createObjectURL(blob);
+          a.href = url;
+
+          var filename = '';
```
Jaydee123, line 183: I would make this 'let' too :)

client/src/pages/spectrogram/components/PluginsPane.tsx
```diff
@@ -94,7 +100,11 @@ export const PluginsPane = ({ cursorsEnabled, handleProcessTime, meta, setMeta }
           toast.error(`Plugin failed to run: ${data.status}`);
           return;
         }
-        if (data.data_output && data.data_output.length > 0) {
+        if (
+          data.data_output[0]['data_type'] == MimeTypes.IQ_SIGMF_32 &&
```
Jaydee123, line 104: I would move to after checking data.data_output is null and data.data_output.length>0

client/src/pages/spectrogram/components/PluginsPane.tsx
sebastus: ok

client/src/pages/spectrogram/components/PluginsPane.tsx
sebastus: good catch

client/src/pages/spectrogram/components/PluginsPane.tsx
sebastus: done

client/src/pages/spectrogram/components/PluginsPane.tsx
```diff
@@ -95,70 +101,98 @@ export const PluginsPane = ({ cursorsEnabled, handleProcessTime, meta, setMeta }
...
-            f.transform(out, samples_slice); // assumes input (2nd arg) is in form IQIQIQIQ and twice the length of fft.size
-
-            out = out.map((x) => x / fftSize); // divide by fftsize
-
-            // convert to magnitude
-            let magnitudes = new Array(out.length / 2);
-            for (let j = 0; j < out.length / 2; j++) {
-              magnitudes[j] = Math.sqrt(Math.pow(out[j * 2], 2) + Math.pow(out[j * 2 + 1], 2)); // take magnitude
-            }
+          if (data.data_output[0]['data_type'] == MimeTypes.IQ_SIGMF_32) {
```
Jaydee123, line 104: you should still be able to do:

        if (data.data_output && data.data_output.length > 0 && data.data_output[0]['data_type'] == MimeTypes.IQ_SIGMF_32) 
        
It's just the ordering of it matters as expressions are executed from left to right

sebastus: I did it this way because of the else. That code also needs to be protected by the null & array element test.

client/src/pages/spectrogram/components/PluginsPane.tsx
```diff
@@ -95,70 +101,98 @@ export const PluginsPane = ({ cursorsEnabled, handleProcessTime, meta, setMeta }
...
+            // just show the first output for now, 99% of plugins will have 0 or 1 IQ output anyway
+            const samples_base64 = data.data_output[0]['samples'];
+            const samples = convertBase64ToFloat32Array(samples_base64);
+            //const sample_rate = data.data_output[0]['sample_rate']; // Hz
```
Nepomuceno, line 108: sould we remove commented out code?

Jaydee123 (approved):
As a future refactoring improvement, I think it would be better to have:

switch (data.data_output[0]['data_type'])

be at the start insted of the 'if else' and then depending on what the type is depends on the functionality we do, IQ_SIGMF_32 code would also go into this, and then the code within there could be split off into different reusable methods

Nepomuceno (approved):
Could do some improvement later by making the treatment of iq files into its own function so we do not have such a big function to treat return

client/src/pages/spectrogram/components/PluginsPane.tsx
sebastus: It's there from before I touched it. Want to turn this task into general cleanup as well?

sebastus:
I agree, that refactoring this code would add a lot of value.

PR #284, opened 2023-06-29
Simple query api

PaulDFoster:
### What does this PR do?
Implementation of a simple query api using query parameters of a GET.
A selection of fields judged to be useful have been implemented, where fields are of type string a case insensitive regex comparison is made.
This list of fields will be refined through discussion with the customer.

### All Submissions:

### Changes to Core Features:


api/tests/test_query_api.py
```diff
@@ -0,0 +1,31 @@
+from unittest.mock import MagicMock
+import database.database
+from main import app
+from .test_data import valid_metadata_array
```
Nepomuceno, line 4: Havent seen thsi before nice way to import

api/handlers/metadata.py
```diff
@@ -121,6 +124,60 @@ async def get_meta_thumbnail(
...
 
 
+@router.get(
+    "/api/datasources/query",
+    status_code=200,
+    response_model=list[Metadata],
+)
+def query_meta(
+    account: Optional[str] = Query(None),
+    container: Optional[str] = Query(None),
```
Nepomuceno, line 134: would be possible for account and containers to be str[] instead of str ? or maybe they are str but we do split by comma or something like that.

Nepomuceno (approved):
LGTM just a small comment if something would be possible because I think it would givce the endpoint more flexibility aprt from that great.

api/handlers/metadata.py
PaulDFoster: Done

PR #301, opened 2023-07-04
enable null blob connection for public blob storage

sebastus:
### What does this PR do?
* add test for null blob connection string, interpret that to be a public account
* change UI so that nothing appears in the SAS token expiration area if public account
* change auth to blob storage so an empty sas token is accepted
* fix the makefile target "run-plugins"
* test empty sas token with the gnuradio/iqengine test
* add test for existing sas token with daniel esteves container

### All Submissions:

### Changes to Core Features:


client/package.json
```diff
@@ -4,6 +4,7 @@
   "private": true,
   "dependencies": {
     "@azure/storage-blob": "^12.14.0",
+    "@azure/identity": "^3.2.3",
```
Nepomuceno, line 7: Is this package needed?

777arc:
Looks great!

777arc (approved):

client/src/pages/repo-browser/RepositoryTile.tsx
```diff
@@ -15,13 +15,27 @@ export const RepositoryTile = (props) => {
   const [isDisabled, setIsDisabled] = useState(false);
   const [isWarning, setIsWarning] = useState(false);
   const [isError, setIsError] = useState(false);
+  const [isPublic, setIsPublic] = useState(false);
```
Nepomuceno, line 18: There is no need for this it awill not be displayed is there is no warning or errors so we can jsut skyp the validation in case there the sas token is empty

client/src/pages/repo-browser/RepositoryTile.tsx
```diff
@@ -15,13 +15,27 @@ export const RepositoryTile = (props) => {
...
   const [dayDifference, setDayDifference] = useState<number>();
   const [expires, setExpires] = useState();
   const [writeableBool, setWriteableBool] = useState<any>();
 
   useEffect(() => {
-    const tempExpires = sasToken.slice(sasToken.search('se')).split('&')[0].slice(3, 13); // YEAR-MONTH-DAY
-    const writeable = sasToken.slice(sasToken.search('sp')).split('&')[0].includes('w'); // boolean
+    let writeable = false;
+    if (sasToken == '') {
+      setIsPublic(true);
```
Nepomuceno, line 26: you could just use if(!!sastoken) and avoid the else

client/package.json
sebastus: good catch

client/src/pages/repo-browser/RepositoryTile.tsx
sebastus: Line 64 - 

        {!isError && !isWarning && (

I am getting the text "SAS Token Expiration: {expires}". I checked the history for the file and didn't see any changes to it, so I don't think it's an error. Adding !IsPublic removes it.

client/src/pages/repo-browser/RepositoryTile.tsx
sebastus: still need IsPublic flag

PR #308, opened 2023-07-04
Make zoom levels index an array

tobybrad:
### What does this PR do?

First stage in increasing the zoom levels available. Rather than use the slider directly we're indexing into an array of zoom levels. At present the zoom levels are 1-10, so no change to functionality right now, but intended to go up to 10-100k. Change following this will make the fetching of tiles performant when we have large zoom levels.

### All Submissions:

* [✅] Have you followed the guidelines in our Contributing document?
* [✅] Have you successfully ran tests with your changes locally? Including container validation.
* [✅] Have you lint your code locally prior to submission?
* [✅] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?
* [✅] Have you labelled this PR correctly? E.G. major, milestone, breaking, breaking-change, minor, feature, enhancement, patch, fix, chore, refactor etc

### Changes to Core Features:

* [✅] Have you added an explanation of what your changes do and why you'd like us to include them?
* [X] Have you written new tests for your core changes, as applicable? (N/A)


client/src/pages/spectrogram/SettingsPane.tsx
```diff
@@ -154,7 +158,7 @@ print("Time elapsed:", (time.time() - start_t)*1e3, "ms")`,
           className="range range-xs range-primary"
           value={state.zoomLevel}
           min={1}
-          max={10}
+          max={zoomLevels.length}
```
Nepomuceno, line 161: Isn't having the minimum here being 1 would would not allow to select the first one in the array and the max you go over the max index of the array

PR #310, opened 2023-07-05
Fix initial loading

Jade-Codes:
### What does this PR do?
On staging there is weird rendering before the page initially loads.
This fixes that.

### All Submissions:


Nepomuceno (approved):
GREAT WORK! 🥳
Where did you get the idea of using this lazy for the routes 😉
I like hos the first print of the home page it is a lot faster now.
We still need to do some work to slim down that spectrogram page

Jade-Codes:
> GREAT WORK! 🥳 Where did you get the idea of using this lazy for the routes 😉 I like hos the first print of the home page it is a lot faster now. We still need to do some work to slim down that spectrogram page

it came to me in a dream

PR #332, opened 2023-07-10
added meta-data-query section

jh87692:
### What does this PR do?

Added the simple query page which interacts with the API server.
The results are a placeholder at the moment as the File component isn't suitable for rendering elements returned.

### All Submissions:

### Changes to Core Features:

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -149,7 +149,7 @@ export const App = () => {
           </div>
           <Link to="/" onClick={() => {}}>
             <div id="IQEngineLogo" className="absolute mt-4 pt-2 top-0 left-1/2 transform -translate-x-1/2 flex">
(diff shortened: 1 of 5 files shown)


Nepomuceno:
Should we change the file names to kebab-case?

jh87692:
> Should we change the file names to kebab-case?

I shall do that now sir!

777arc:
good point that the File/Directory components arent set up to render elements returned, but if you look at the metadata_collection.data structure (see below), I think we can have the results of the query essentially replace that structure
image

jh87692:
> good point that the File/Directory components arent set up to render elements returned, but if you look at the metadata_collection.data structure (see below), I think we can have the results of the query essentially replace that structure image

how would we like the results displaying?

777arc:
We can do it as a separate PR imo

PR #334, opened 2023-07-10
Query API update for geolocation fields

PaulDFoster:
### What does this PR do?
A small modification to the parameters for geolocation in the query API.
Geolocation is now available for the geolocation field in captures and annotations. 
Geo parameters are:

annotations_geo and captures_geo

annotations_geo or captures_geo - 
Specifies the geographic location to search within for 'captures.core:geolocation' or 'annotations.core:geolocation'. 
Format: 'longitude,latitude,max_radius'. 
Longitude and latitude define the centre point of the search area, and max_radius (in meters) defines the search radius around this point. 
              Example: 'captures_geo=-175.8,4.4,100000' performs a search within a 100000 meter radius around the point (-175.8, 4.4).

### All Submissions:

* [] Have you lint your code locally prior to submission?

### Changes to Core Features:


api/handlers/metadata.py
```diff
@@ -178,6 +178,34 @@ async def get_meta_thumbnail(
     )
 
 
+def process_geolocation(target: str, geolocation: str):
```
Nepomuceno, line 181: Wouldn't be better to receive a geometry here directly instead of a comma separated field?

api/handlers/metadata.py
```diff
@@ -195,9 +223,8 @@ def query_meta(
     min_datetime: Optional[datetime] = Query(None),
     max_datetime: Optional[datetime] = Query(None),
     text: Optional[str] = Query(None),
-    geo_lat: Optional[float] = Query(None),
-    geo_long: Optional[float] = Query(None),
-    geo_radius: Optional[float] = Query(None),
+    captures_geo: Optional[str] = Query(None),
```
Nepomuceno, line 226: I would.kist make this geometry instead of comma separated strings

Nepomuceno (approved):
Changing from the comma separated string to a geometry gives more flexibility for the method but the rest looks really good

api/handlers/metadata.py
PaulDFoster: Agree geometry is ultimate flexibility. However, its verbose as a query parameter, it requires more client side code/understanding, and we only have two geolocation fields at present.
The requirements from the customer are for two searches sensor location and transmission location. These are directly addressed at this time in this code as captures and annotations.

PR #339, opened 2023-07-11
Query API datetime fix

PaulDFoster:
### What does this PR do?
query_condition.update did not perform as expected, replacing the condition for captures.core:datetime rather than adding to it with a subsequent update. The error was missed because of insufficient test data, and discovered using a large dataset.
The fix is to construct the range and then apply it to the captures.core:datatime field.

### All Submissions:

### Changes to Core Features:


Nepomuceno (approved):
Nice had never used the update function of a ductionary before.

PR #362, opened 2023-07-18
Admin page - Adding pluginEndpoint

Nepomuceno:
### What does this PR do?

Add admin page for plugin

### All Submissions:


client/src/api/plugin/queries.tsx
```diff
@@ -0,0 +1,150 @@
...
+    }
+  );
+}
+
+export function useGetPluginDetailed(plugin: PluginDefinition) {
+  return useQuery<PluginEndpoint>(
+    ['plugin', plugin.name, 'detailed'],
+    async () => {
+      // make sure we have a trailing slash  to concatenate the url
+      if (plugin.url && !plugin.url.endsWith('/')) plugin.url += '/';
```
Jade-Codes, line 59: This feels very wrong

client/src/api/plugin/queries.tsx
```diff
@@ -0,0 +1,150 @@
...
+      const response = await axios.get<string[]>(plugin.url);
+      if (!response.data) throw new Error(`Plugin ${plugin.name} does not return correctly`);
+      const result = new PluginEndpoint();
+      result.name = plugin.name;
+      result.url = plugin.url;
+      result.plugins = {};
+      for (const item of response.data) {
+        const pluginUrl = new URL(item, plugin.url);
+        const pluginParameters = await axios.get<PluginParameters>(pluginUrl.toString()).catch((err) => {
```
Jade-Codes, line 68: should we really be doing this at this point? surely it would be better to only do this when we open the modal?

client/src/api/plugin/queries.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+export function useGetPluginDetailed(plugin: PluginDefinition) {
+  return useQuery<PluginEndpoint>(
+    ['plugin', plugin.name, 'detailed'],
+    async () => {
+      // make sure we have a trailing slash  to concatenate the url
+      if (plugin.url && !plugin.url.endsWith('/')) plugin.url += '/';
+      const response = await axios.get<string[]>(plugin.url);
+      if (!response.data) throw new Error(`Plugin ${plugin.name} does not return correctly`);
+      const result = new PluginEndpoint();
```
Jade-Codes, line 62: can we not just do new PluginEndpoint { ... }?

client/src/api/plugin/queries.tsx
Jade-Codes, line 68: can pluginUrl ever return null?

client/src/api/plugin/queries.tsx
```diff
@@ -0,0 +1,150 @@
...
+      result.name = plugin.name;
+      result.url = plugin.url;
+      result.plugins = {};
+      for (const item of response.data) {
+        const pluginUrl = new URL(item, plugin.url);
+        const pluginParameters = await axios.get<PluginParameters>(pluginUrl.toString()).catch((err) => {
+          console.error(`Failing getting plugin parameters from ${pluginUrl}`, err);
+          return { data: null };
+        });
+        if (!pluginParameters.data) {
```
Jade-Codes, line 72: could plugin params ever be null?

client/src/pages/admin/pages/plugins.tsx
```diff
@@ -1,11 +1,248 @@
-import React from 'react';
+import { PluginDefinition } from '@/api/Models';
```
Jade-Codes, line 1: I think you need to split this into the Add/Create/List components

client/src/api/plugin/queries.tsx
Nepomuceno: I know I did feel the dame them I looked at the result of this.
What do you think it is the result of this
new URL("test",  "https://plugin.com/plugins")
and what is the result os this other function
new URL("test", "https://plugin.com/plugins/")

a tip... they are not equal.

client/src/api/plugin/queries.tsx
Nepomuceno: We could have another query that would do this when we do open the modal actually we could have this query as part of the modal since prior to the modal no information in this query it is needed.

Jade-Codes: just thinking that it's better to not fetch the data before is needed

client/src/api/plugin/queries.tsx
Nepomuceno: Unfortunetly this does not work in js I could do this way though do you think it is better ?
```js
const result = {
        name: plugin.name,
        url: plugin.url,
        plugins: {},
      } as PluginEndpoint;
```

client/src/api/plugin/queries.tsx
Nepomuceno: it is a required field of a plugin.

client/src/api/plugin/queries.tsx
Nepomuceno: If plugin parrameters it is null it would be catched by the axios error when trying to convert that to a plugin parameters object.

client/src/pages/admin/pages/plugins.tsx
Nepomuceno: I think there is no need to create one file per component  but can split the create and add into their own components but I would keep them in this same file for now until we think there is reason to split I would say. 

https://www.youtube.com/watch?v=nYNn_sNcBbM

client/src/api/plugin/queries.tsx
Nepomuceno: Agreed will change the query to just be enabled is the modal is opened that way would avoid to fetch before it is needed

Nepomuceno: Changed

client/src/pages/admin/pages/plugins.tsx
Jade-Codes: yep much more readable thanks

client/src/pages/admin/pages/__tests__/plugins.test.tsx
```diff
@@ -1,12 +1,117 @@
-import { describe, expect, test } from 'vitest';
-import { render, screen } from '@testing-library/react';
+import { beforeEach, describe, expect, test } from 'vitest';
+import { render, screen, waitFor } from '@testing-library/react';
 import Plugins from '@/pages/admin/pages/plugins';
 import '@testing-library/jest-dom';
+import userEvent from '@testing-library/user-event';
 import React from 'react';
+import { useAllProviders } from '@/mocks/setup-tests';
+import nock from 'nock';
```
Jade-Codes, line 9: do we need some tests for the edit, add and delete?

client/src/api/plugin/queries.tsx
Jade-Codes: I think that is better

Jade-Codes (approved):

client/src/api/plugin/queries.tsx
```diff
@@ -0,0 +1,150 @@
...
+    }
+  );
+}
+
+export function useGetPlugin(plugin: PluginDefinition) {
+  return useQuery<string[]>(
+    ['plugin', plugin.name],
+    async () => {
+      const response = await axios.get<string[]>(plugin.url);
+      // make sure we onlyt return in case of a string array
```
hannahlspencer, line 31: Picky comment but typo - only

client/src/pages/admin/pages/__tests__/configuration.test.tsx
```diff
@@ -144,13 +143,7 @@ describe('Test Configuration', () => {
 
     nock('http://localhost:3000').put(`/api/config`).reply(200);
 
-    render(
-      <>
-        <Toaster />
```
hannahlspencer, line 149: How come you're removing Toaster from these tests, does it work OK without?

hannahlspencer (approved):
Couple of Qs but LGTM!

client/src/pages/admin/pages/__tests__/configuration.test.tsx
Nepomuceno: I did add toaster to the main wrapper that way the main wrapper will have it for every test to use

PR #365, opened 2023-07-18
add use get image hook

sebastus:
### What does this PR do?

add useGetImage custom hook
add several tests for the rgb generator in selector.ts
add mock component

### All Submissions:

### Changes to Core Features:


client/src/pages/spectrogram/hooks/__tests__/use-get-image.test.tsx
```diff
@@ -0,0 +1,152 @@
...
+    newFftData[opIdx + 1] = colMaps[colorMap][sigVal][1]; // green
+    newFftData[opIdx + 2] = colMaps[colorMap][sigVal][2]; // blue
+    newFftData[opIdx + 3] = 255; // alpha
+  }
+  expectedImageData = new ImageData(newFftData, fftSize, num_ffts);
+
+  return { sampleRecording, num_ffts, expectedImageData };
+};
+
+const imagesAreTheSame = (imageDataA: ImageData, imageDataB: ImageData) => {
```
Jade-Codes, line 71: I would generally try to avoid having logic like this within a test, I would split out the tests cases for when successful and when non successful and remove the if/else logic

client/vitest.setup.ts
```diff
@@ -0,0 +1 @@
+import 'vitest-canvas-mock';
```
777arc, line 1: is there a way to avoid needing this extra file?

sebastus: no - it provides the capability to extract the imageData from the imageBitmap that comes out of fftToRGB, so it can be compared to the expected outcome in tests.

sebastus: though maybe as a dev dependency?

client/src/pages/spectrogram/hooks/__tests__/use-get-image.test.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+const normalizeMagnitude = (magnitude_in_db: number, magnitudeMin: number, magnitudeMax: number) => {
+  const dbPer1 = 255 / (magnitudeMax - magnitudeMin);
+  let magnitude = (magnitude_in_db - magnitudeMin) * dbPer1; // normalize to 0-255
+  magnitude = magnitude > 255 ? 255 : magnitude; // clip above 255
+  magnitude = magnitude < 0 ? 0 : magnitude; // clip below 0
+  return magnitude;
+};
+
+const generateSampleRecording = (
```
Jade-Codes, line 25: I would move this to a function within the utilities folder

client/src/pages/spectrogram/hooks/__tests__/use-get-image.test.tsx
```diff
@@ -0,0 +1,152 @@
...
+const BucketsOfDb = [-100, -90, -80, -70, -60, -50, -40, -30, -20, -10, 0, 10, 20, 30, 40, 50];
+
+// types of sample recordings
+enum SampleType {
+  MultipleBuckets = 0,
+  WhiteBox = 1,
+}
+
+const normalizeMagnitude = (magnitude_in_db: number, magnitudeMin: number, magnitudeMax: number) => {
+  const dbPer1 = 255 / (magnitudeMax - magnitudeMin);
```
Jade-Codes, line 18: same as the previous

client/vitest.setup.ts
sebastus: just confirmed - it's only installed during test.

777arc: sorry i meant the file itself, is there a way to include it in vite.config.ts without needing a whole separate file?

client/src/pages/spectrogram/hooks/__tests__/use-get-image.test.tsx
sebastus: done

client/vitest.setup.ts
Nepomuceno: We might change the place so it is not another file in root.

client/src/utils/testFunctions.ts
```diff
@@ -0,0 +1,64 @@
+import { colMaps } from '@/utils/colormap';
```
Jade-Codes, line 1: are these test functions? I'd argue 'normaliseMagnitude' and 'generateSampleRecording' could be used for multiple purposes

client/vitest.setup.ts
sebastus: removed the file. we'll see if it's really needed.

client/src/utils/testFunctions.ts
sebastus: they are used only for test functions. They may be used for something else later - no way to know.

Nepomuceno:
Not a big change but would be good to look at some tests with smaller fft numbers to like 128 , 256 and 512

client/src/utils/testFunctions.ts
Nepomuceno: I agree about normaliseMagnitude ( but this puts in question the whole validity of the tests ( I mean having some fixed values to tests with that function would make sense )

Nepomuceno: I think the best option to test this it is to use well known values for normalisation instead of a function.

sebastus: I prefer Jade's solution. It enables more tests.

client/src/pages/spectrogram/hooks/__tests__/use-get-image.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+    ['256,white box', tilesize, 256, -30.0, -20.0, 'jet', SampleType.MultipleBuckets],
+    ['512,white box', tilesize, 512, -60.0, 10.0, 'plasma', SampleType.MultipleBuckets],
+    ['defaults,white box', tilesize, 1024, -40.0, -10.0, COLORMAP_DEFAULT, SampleType.MultipleBuckets],
+    ['defaults,jet,white box', tilesize, 1024, -40.0, -10.0, 'jet', SampleType.MultipleBuckets],
+    ['2048,jet,white box', tilesize, 2048, -40.0, -10.0, 'jet', SampleType.MultipleBuckets],
+    ['-80/0,plasma,white box', tilesize, 1024, -80.0, 10.0, 'plasma', SampleType.MultipleBuckets],
+    ['4096,-20/20,white box', tilesize, 4096, -20.0, 20.0, 'plasma', SampleType.MultipleBuckets],
+    ['8192,-60/50,inferno,white box', tilesize, 8192, -60.0, 50.0, 'inferno', SampleType.MultipleBuckets],
+    ['16384,-100/50,white box', tilesize, 16384, -100.0, 50.0, 'inferno', SampleType.MultipleBuckets],
+  ])('Passing test: %s', async (comment, tile_size, fftSize, magnitudeMin, magnitudeMax, colorMap, sampleType) => {
```
Jade-Codes, line 80: Is there a way we could make this better named?

sebastus: updated

Nepomuceno (approved):
LGTM

Jade-Codes (approved):

PR #370, opened 2023-07-19
Create IQEngine meta query page

Nepomuceno:
### What does this PR do?

Create query engine query page.

image

image

### All Submissions:


client/src/hooks/use-router.tsx
```diff
@@ -76,6 +83,13 @@ export function useRouter() {
                 return { Component: Configuration };
               },
             },
+            {
+              path: 'query',
```
Jade-Codes, line 87: Why added twice?

client/src/pages/metadata-query/__tests__/date-query.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+        description={mockDescription}
+        validator={validatorMock}
+        handleQueryValid={handleQueryValidMock}
+        handleQueryInvalid={handleQueryInvalidMock}
+      />
+    );
+    const dateFrom = screen.getByTestId('date-from');
+    const dateTo = screen.getByTestId('date-to');
+
+    fireEvent.change(dateFrom, { target: { value: '2023-01-01' } });
```
Jade-Codes, line 66: It is berrer to use a user event

client/src/pages/metadata-query/__tests__/freq-query.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+      to: 300000000,
+    });
+    const upperBound = screen.getByTestId('freq-upper');
+    fireEvent.change(upperBound, { target: { value: '1' } });
+    expect(validatorMock).toHaveBeenCalledWith({
+      from: 30000000,
+      to: '1',
+    });
+    expect(handleQueryInvalidMock).toHaveBeenCalledTimes(1);
+  })
```
Jade-Codes, line 78: Pls no

client/src/pages/metadata-query/__tests__/geo-query-type.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+      />
+    );
+    const type = screen.getByLabelText('captures');
+    fireEvent(type, new MouseEvent('click', {
+      bubbles: true,
+      cancelable: true,
+    }));
+    expect(mockHandle).toHaveBeenCalledTimes(1);
+    expect(mockHandle).toHaveBeenCalledWith('captures');
+  });
```
Jade-Codes, line 28: Pls no

client/src/pages/metadata-query/freq-query.tsx
```diff
@@ -0,0 +1,106 @@
...
+          </button>
+        </div>
+      </div>
+      {show && (
+        <div className="card bg-neutral text-neutral-content">
+          <div className="card-body">
+            <BandSelection selected={band[0]} handleSelection={handleSelection} bands={bands} />
+            <h3 className="text-lg">
+              From:
+              <span className="ml-2 badge badge-md badge-success">{`${freqRange.from} Hz`}</span>
```
Jade-Codes, line 71: Why not use the dual range slider?

client/src/pages/metadata-query/metadata-query.tsx
```diff
@@ -0,0 +1,113 @@
...
+  return (
+    <div className="m-10 mt-100">
+      <h1 className="text-3xl font-bold">Field Selection</h1>
+      <div className="grid grid-cols-10 gap-3">
+        <div className="col-span-1">
+          <div className="form-control">{renderQuerySelection()}</div>
+        </div>
+        <div className="col-span-9 ml-10 ">
+          {renderQueryComponents()}
+          <button onClick={handleQuery} disabled={!showQueryButton()} className="btn btn-secondary w-full">
```
Jade-Codes, line 103: Can we make button styling consistent with other pages

client/src/pages/metadata-query/__tests__/freq-query.test.tsx
Nepomuceno: Couold you provide any reason whhy or a viable alternative ?

client/src/pages/metadata-query/__tests__/geo-query-type.test.tsx
Nepomuceno: Couold you provide any reason whhy or a viable alternative ?

client/src/pages/metadata-query/freq-query.tsx
Nepomuceno: This could definetly be improved to a dual range slider but in order to not make this pr even bigger than it already is I would add this as an improvement we could do later on.

client/src/pages/metadata-query/__tests__/geo-query-type.test.tsx
Jade-Codes: How does calling a method prove the functionality we want is done. Test the return you see in the page rather than that. I.e is the button still disabled vs is it not.

client/src/pages/metadata-query/__tests__/freq-query.test.tsx
Jade-Codes: How does calling a method prove the functionality we want is done. Test the return you see in the page rather than that. I.e is the button still disabled vs is it not

client/src/pages/metadata-query/metadata-query.tsx
Nepomuceno: Removed the styling to keep consistency..

client/src/pages/metadata-query/__tests__/freq-query.test.tsx
Nepomuceno: The problem it is that for these componenets there is no screen information for this method the only thing that this does it is to return the value to the query page that them use those values to compose the query because of that testing tdirect the method that have been called we could test if the method was called with the right arguments like I chamged in the date query. Idealy this component could be done as a hook that way we could subscribe top the state that is stored and by that we could inspect the state that is done inside this object instead of proping up like that.

client/src/pages/metadata-query/__tests__/geo-query-type.test.tsx
Nepomuceno: This is the same case of the other methos thic component in isolation does not provide impediate feedback for any change in this method. it only makes a callback to to object that did mount that.

client/src/pages/metadata-query/__tests__/date-query.test.tsx
Nepomuceno: Changed to use userEvent

client/src/pages/metadata-query/__tests__/freq-query.test.tsx
Nepomuceno: But this would not be a test of this component this component knows nothing about where it is used it has no knowledge of the query page at all it doesn't even know that one exists.

I agree that a test on the query page that does validate that in case you input a valid date the query button is enable but them if you put an invalid date it is disabled would be useful but that was not the intent of this test file the intent of this test file was to test this component in isolation outside of the context of the query page.

We might get to the conclusion that there is no need to test the component in isolation. So this test will be deemed as obsolete and could be removed.

client/src/hooks/use-router.tsx
Nepomuceno: Removed the extr add was probably a merge mistake

777arc:
Screenshot?  Where will this component be located, top of recordingslist page?

hannahlspencer (approved):

PR #378, opened 2023-07-21
Extract modal

hannahlspencer:
### What does this PR do?
- Extracts out existing modals so we have one reusable modal element.
- Refactors add data source so it's a modal rather than its own page

### All Submissions:


Nepomuceno:
One thing that we would need to be able to pass in it is the handle submit action to be added to the form

hannahlspencer:
> One thing that we would need to be able to pass in it is the handle submit action to be added to the form

The handleSubmit is within the DataSourceForm, where else would it need to be passed?

Nepomuceno:
> > One thing that we would need to be able to pass in it is the handle submit action to be added to the form
> 
> The handleSubmit is within the DataSourceForm, where else would it need to be passed?

Modal has a form on it you should be able to pass which action you want to happen when you submit that form.

hannahlspencer:
> > > One thing that we would need to be able to pass in it is the handle submit action to be added to the form
> > 
> > 
> > The handleSubmit is within the DataSourceForm, where else would it need to be passed?
> 
> Modal has a form on it you should be able to pass which action you want to happen when you submit that form.

Only the Modal the DataSourceForm is using, not every Modal that will use the element has a submit button

client/src/pages/spectrogram/components/annotation/__tests__/Actions.test.tsx
```diff
@@ -196,8 +199,8 @@ describe('Annotation list component', () => {
     await userEvent.click(updateButton);
 
     // Assert
-    const modal = await screen.findByLabelText('Annotation 0 Modal');
-    expect(modal).toHaveClass('modal-open');
+    const modal = await screen.queryByLabelText('Modal Dialog');
+    expect(modal).toBeInTheDocument();
```
Jade-Codes, line 203: does toBeVisible() pass?

client/src/pages/admin/pages/add-data-source.tsx
```diff
@@ -50,49 +50,47 @@ export const DataSourceForm = () => {
...
       return (
-        <div>
-        <h1>Add data source</h1>
         <form ref={formRef} onSubmit={handleSubmit}>
         <input
             type="text"
             name="name"
             placeholder="Data Source Name"
-            className="input input-bordered input-success w-full max-w-xs"
+            className="input input-bordered input-success w-full max-w-xs text-white"
```
Jade-Codes, line 58: I would change the input colour to be the opposite of the input background, so if your background colour is bg-base-100 it should be bg-base-content: https://daisyui.com/docs/colors/

What does the .input class set as default bg colour?

client/src/pages/spectrogram/components/annotation/__tests__/Actions.test.tsx
hannahlspencer: Good point actually, this shouldn't be visible in this test as the update should close it. Will remove!

client/src/pages/spectrogram/components/annotation/__tests__/Actions.test.tsx
```diff
@@ -196,9 +199,6 @@ describe('Annotation list component', () => {
     await userEvent.click(updateButton);
 
     // Assert
```
Jade-Codes, line 201: We should still be checking the update button is disabled

client/src/pages/spectrogram/components/annotation/__tests__/Actions.test.tsx
```diff
@@ -159,8 +162,8 @@ describe('Annotation list component', () => {
...
-    const modal = await screen.findByLabelText('Annotation 0 Modal');
-    expect(modal).toHaveClass('modal-open');
+    const modal = await screen.queryByLabelText('Modal Dialog');
+    expect(modal).toBeInTheDocument();
```
Jade-Codes, line 166: Is toBeVisible working here?

client/src/features/ui/modal/__tests__/Modal.test.tsx
Jade-Codes, line 1: I would change this so we aren't testing a component we have just created, instead I would have two tests that test the ModalDialog directly
- Set show modal is passed through with a true parameter
- Set show modal is passed through with a false parameter

Then I would test based on those that the modal is either visible or not.

client/src/pages/admin/pages/plugins.tsx
```diff
@@ -244,33 +196,30 @@ export const PluginAdd = () => {
...
             </button>
           </div>
         </form>
-      </Modal>
     </>
   );
 };
 
 export const PluginRow = ({ plugin, removePlugin }: PluginRowProps) => {
+  const [showModal, setShowModal] = useState(false);
```
Jade-Codes, line 209: Is this being used for both dialogs? Will that not make both dialogs show when clicking either 'edit' or 'add' button?

Jade-Codes: Nvm I can see these are separate methods, I would change the name so you associate it to the edit button, so it's showEditModal, setShowEditModal for example

client/src/features/ui/modal/Modal.tsx
```diff
@@ -0,0 +1,37 @@
...
+      useOnClickOutside(ref, () => {
+        setShowModal(false)
+      });
+
+      return (
+        <dialog aria-label={'Modal Dialog'} className="modal modal-open w-full h-full">
+            <div className="modal-box">
+                <h3 className="font-bold text-lg text-primary">{heading}</h3>
+                <button
+                    aria-label={'Close Button'}
```
Jade-Codes, line 25: I need to go back and rename some aria labels, but they don't need the 'Button or Dialog' in the name, would you be able to change it for the ones in this file?
https://www.aditus.io/aria/aria-label/

client/src/pages/spectrogram/components/annotation/__tests__/Actions.test.tsx
hannahlspencer: It doesn't, but I have confirmed that if the modal isn't open then toBeInTheDocument fails

Jade-Codes (approved):
I really like the refactoring of the modal into it's own component, good work!

PR #379, opened 2023-07-22
Auto loading data source

Nepomuceno:
### What does this PR do?

Automatically loading data source from environment variables.

This would make the time that the api is standup it would automatically load to the current database all the datasources that are in the current environment variable for connections.

### All Submissions:


api/database/datasource_repo.py
```diff
@@ -51,3 +52,31 @@ async def datasource_exists(account, container) -> bool:
...
+
+    Returns
+    -------
+    DataSource
+        The datasource.
+    """
+    datasource_collection: AgnosticCollection = collection()
+    if await datasource_exists(datasource.account, datasource.container):
+        raise Exception("Datasource Already Exists")
+    if datasource.sasToken:
```
hannahlspencer, line 76: Really like this extraction - is there any potential future need to do this with the update as well?

Nepomuceno: I do think it would make sense to add this to others to avoid keeping the logic in the route itself but I would say one thing at time.

hannahlspencer (approved):
LGTM!

PaulDFoster (approved):

PR #381, opened 2023-07-25
New data load

Nepomuceno:
### What does this PR do?

This PR introduces several enhancements related to data loading methods for printing an image on the screen using data obtained from the API or Blob storage.

**Things this PR does:**

- Adds cancel signal to all IQ methods. This ensures that when a query is no longer being observed, it will not attempt to fetch the data. This improvement prevents unnecessary data fetching and addresses an issue where fast scrolling would trigger data requests for elements that are no longer visible on the screen. This problem was present in the previous implementation but became more noticeable with the ability to make requests even for just one FFT worth of data in the case of scrolling.
    
- Introduces the capability to download data in blocks and indexes. Users can specify the size of the IQ data block they want and the specific blocks they want to download.
    
- Implements a hook that calculates the required data to be downloaded based on the current spectrogram parameters, taking into account the data that has already been downloaded.
    
- Adds cache functionality to the hook, ensuring that only tiles that have not been seen before are downloaded. This optimization further improves data loading efficiency.

### All Submissions:


client/src/utils/group.ts
```diff
@@ -0,0 +1,15 @@
+export function groupContingousIndexes(indexes: number[]) {
```
Jade-Codes, line 1: might be worth adding the same test cases I added to the python

client/src/api/iqdata/Queries.ts
```diff
@@ -140,3 +140,72 @@ export const useCurrentCachedIQDataSlice = (meta: SigMFMetadata, tileSize: numbe
...
+      const iqDataClient = IQDataClientFactory(type, filesQuery.data, dataSourcesQuery.data);
+      const iqData = await iqDataClient.getIQDataBlocks(meta, fftsRequired, fftSize, signal);
+      console.debug('useGetIQData', iqData);
+      return iqData;
+    },
+    enabled: !!meta && !!filesQuery.data && !!dataSourcesQuery.data,
+  });
+
+  useEffect(() => {
+    console.debug('iqQuery Status', status, error);
```
Jade-Codes, line 173: is this needed

.gitignore
```diff
@@ -50,3 +50,4 @@ tiling/spectrogram.pmtiles
 
 # in case someone by mistake do npm i in the root do not check in package-lock.json
 /package-lock.json
+dev-page.tsx
```
Jade-Codes, line 53: ?

client/src/utils/group.ts
Nepomuceno: Will do.

client/src/api/iqdata/Queries.ts
Nepomuceno: Removed it was only for debug

.gitignore
Nepomuceno: Removed was there for debug

client/src/utils/group.ts
Nepomuceno: Added tests 👍

client/src/api/iqdata/BlobClient.ts
```diff
@@ -3,34 +3,91 @@ import { convertToFloat32 } from '@/utils/FetchMoreDataSource';
...
+  async getIQDataBlocks(
+    meta: SigMFMetadata,
+    indexes: number[],
+    blockSize: number,
+    signal: AbortSignal
+  ): Promise<IQDataSlice[]> {
+    console.debug('getIQDataBlocks', indexes);
+    const contingousIndexes = groupContingousIndexes(indexes);
+    let { account, container, file_path } = meta.getOrigin();
+    // if filePath does not finish in .sigmf-data, add it
```
hannahlspencer, line 24: Could there be any reason a file doesn't end with .sigmf-data because it isn't that, or would a check be unnecessary?

client/src/api/iqdata/LocalClient.ts
```diff
@@ -10,11 +10,24 @@ export class LocalClient implements IQDataClient {
...
+    meta: SigMFMetadata,
+    indexes: number[],
+    tileSize: number,
+    signal: AbortSignal
+  ): Promise<IQDataSlice[]> {
+    return Promise.all(indexes.map((index) => this.getIQDataSlice(meta, index, tileSize, signal)));
   }
 
-  async getIQDataSlice(meta: SigMFMetadata, index: number, tileSize: number): Promise<IQDataSlice> {
+  async getIQDataSlice(meta: SigMFMetadata, index: number, tileSize: number, sinal: AbortSignal): Promise<IQDataSlice> {
```
hannahlspencer, line 30: typo in signal

client/src/api/iqdata/BlobClient.ts
Nepomuceno: It would not normally come with .sigmf-data because trhe way we normally display the filename it is without the extrension because the same file name it is used for thumbnail metadata and iqdata

client/src/api/iqdata/LocalClient.ts
Nepomuceno: Fixed

hannahlspencer (approved):

client/src/utils/__tests__/group.test.ts
```diff
@@ -0,0 +1,25 @@
+import { groupContingousIndexes } from '@/utils/group';
```
PaulDFoster, line 1: Contiguous* spelling

client/src/utils/group.ts
```diff
@@ -0,0 +1,22 @@
...
```
PaulDFoster, line 1: Contiguous* spelling

client/src/utils/__tests__/group.test.ts
Nepomuceno: Fixed

PaulDFoster (approved):

PR #385, opened 2023-07-26
Initial server side

Jade-Codes:
### What does this PR do?
Adds the initial server rendering for the returning of iq data

### All Submissions:


api/rf/samples.py
```diff
@@ -30,8 +30,31 @@ def get_samples(data_bytes, data_type) -> np.ndarray:
...
+    if data_type == "cf32_le" or data_type == "cf32" or data_type == "cf32_be":
+        return 8
+    elif data_type == "ci16_le" or data_type == "ci16" or data_type == "ci16_be":
+        return 4
+    elif data_type == "ci8" or data_type == "i8":
+        return 2
+    else:
+        raise ValueError("Datatype " + data_type + " not implemented")
+
+def get_sample_length_from_byte_length(byte_length: int, data_type: str):
```
777arc, line 59: the name of this function is a little ambiguous, but i dont think it has to be a function at all, its just dividing two numbers

api/handlers/iq.py
```diff
@@ -34,6 +38,49 @@ async def get_sas_token(
...
+    
+
+@router.get(
+    "/api/datasources/{account}/{container}/{filepath:path}/iq-data",
+    status_code=200
+)
+async def get_iq_data(
+    filepath: str,
+    fft_arr_str: str,
+    fft_size: int,
```
Nepomuceno, line 47:
```suggestion
    block_indexes: str,
    block_size: int,
```
We are not actaully processing ffts here so changing this to block would avoid confusion

api/rf/samples.py
Nepomuceno: I did create this function it was on a branch where `get_bytes_per_iq_sample` did not exist so it might not make sense now anymore

api/handlers/iq.py
Jade-Codes: Changed

api/rf/samples.py
Jade-Codes: Removed and updated

api/blob/azure_client.py
```diff
@@ -61,6 +61,11 @@ async def get_blob_properties(self, filepath) -> BlobProperties:
         blob_client = self.get_blob_client(filepath)
         return await blob_client.get_blob_properties()
 
+    async def get_blob_size(self, filepath) -> int:
+        blob_client = self.get_blob_client(filepath)
```
Nepomuceno, line 65: Doesn't this method exists?

Jade-Codes: You mean the one YOU COPIED from me ;)

Nepomuceno: Nice!

Nepomuceno (approved):
LOKTM 😜

PR #399, opened 2023-07-27
Add datasource to query

hannahlspencer:
### What does this PR do?
Adds functionality to query the metadata by datasource.

### All Submissions:


client/src/pages/metadata-query/data-source-query.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+export const SourceQuery = ({ validator, queryName, handleQueryValid, handleQueryInvalid }) => {
+  const { apiQuery, blobQuery } = useGetDatasources();
+  const [selections, setSelections] = useState({});
+
+  useEffect(() => {
+    if (apiQuery.data || blobQuery.data) {
+      setDataSources();
+    }
+  }, []);
```
Nepomuceno, line 12: Instead of having an empty dependency array woudl make more sense to have a dependency on apiquery.data and blobquery.data

client/src/pages/metadata-query/data-source-query.tsx
```diff
@@ -0,0 +1,106 @@
...
+    blobQuery.data?.forEach((item) => {
+      dataSources[item.name] = { name: item.name, account: item.account, container: item.container, selected: false };
+    });
+
+    apiQuery.data?.forEach((item) => {
+      dataSources[item.name] = { name: item.name, account: item.account, container: item.container, selected: false };
+    });
+
+    setSelections(dataSources);
+  };
```
Nepomuceno, line 31: This could be inside the other use effect directly no need to create a new function.

Nepomuceno (commented):
I would not do the datasources as just another field in the queries I think they go beyound that. they should be something that gets all selected bby default. I  dont know it doesn't feel for me as like just another field of the queries but the sources upon which you will need to execute your query against. Also I dont think it does make sense to include Blob data sources since you cannot use the query engine to query them.

client/src/pages/metadata-query/data-source-query.tsx
hannahlspencer: I initially had it as that, but this use effect was initialises the data sources to all be checked as false. When I had the data in the dependency array they were set to being unchecked on every re-render, so this empty array makes it so they're only set as unchecked on the first render

hannahlspencer:
> I would not do the datasources as just another field in the queries I think they go beyound that. they should be something that gets all selected bby default. I dont know it doesn't feel for me as like just another field of the queries but the sources upon which you will need to execute your query against. Also I dont think it does make sense to include Blob data sources since you cannot use the query engine to query them.

Afaik they do all get selected by default if they're not chosen as there's no filtering by database_id in the query if that doesn't get selected. Agree with what you mean about the datasources being things that get queried against, but if not on the query page then where? If RFDX is a big data store with lots of little data stores in I think that if you want to only choose some specific data stores then that works as a query. 

Also, agree with blob storage - that was left in as it was how I was testing it working, have now removed.

client/src/pages/metadata-query/data-source-query.tsx
Nepomuceno: The problem with this approach it is that you might start with them empty but them the result from the query might come in and what about them now ? You will not update the page to have the actual values.

hannahlspencer: What would you suggest? Having the dependency on apiquery.data breaks it, are there other options?

Nepomuceno (approved):

777arc (approved):

PR #412, opened 2023-08-02
Add reshape method

tobybrad:
### What does this PR do?

Adds the function:

```
export function reshapeFFTs(currentFFTSize: number, currentData: Float32Array[], newFFTSize: number): Float32Array[]
```
which takes in an array of Float32Array[currentFFTSize] and returns a reshaped array of Float32Array[newFFTSize].

If currentFFTSize>newFFTSize then each element of the original array will create multiple new lines in the output. If currentFFTSize<newFFTSize then multiple elements of the original array will comprise a single element of the output array.
If currentFFTSize == newFFTSize then the original array is returned unchanged.

If, for any reason, there are missing lines in the original array, NaN is used to indicate their absence in the output array.

No attempt is made to resize arrays where currentFFTSize and newFFTSize are not integer multiples of each other.

I believe the code to be comprehensively tested.

### All Submissions:

### Changes to Core Features:


client/src/api/iqdata/Queries.ts
```diff
@@ -200,3 +201,50 @@ export function useGetIQData(type: string, account: string, container: string, f
     setFFTsRequired,
   };
 }
+
+export function reshapeFFTs(currentFFTSize: number, currentData: Float32Array[], newFFTSize: number): Float32Array[] {
```
Nepomuceno, line 205: There is a file called rfFunctions in utils that might be a better fit to where put this function ( that ile also need to be renamed for the new naming convention to rf-functions.ts 

Not a big deal and nothing that I dont think that would hole up the merge.

Nepomuceno (approved):
Apart from movin the function to utils the implementation itself and the tests do look good.

client/src/api/iqdata/__tests__/Queries.test.tsx
```diff
@@ -103,3 +104,79 @@ describe('Check if query cache works correctly ', () => {
...
+    const testArray = createTestArray(fftSize, 800);
+    const newArray = reshapeFFTs(fftSize, testArray, fftSize);
+    expect(newArray).toBe(testArray);
+  });
+
+  test('Reshape to larger fftSize', async ({ expect }) => {
+    const fftSize = 128;
+    const testArray = createTestArray(fftSize, 800);
+    const newArray = reshapeFFTs(fftSize, testArray, fftSize * 2);
+    console.log();
```
hannahlspencer, line 133: why the empty console logs?

hannahlspencer (approved):
LGTM just a query

client/src/api/iqdata/Queries.ts
777arc: Test is in this dir though so should be fine here

PR #418, opened 2023-08-02
Iqdata performance

Jade-Codes:
What does this PR do?
Speeds up the new spectrogram page

* [x ] Have you lint your code locally prior to submission?

### DIFF
--- a/api/handlers/iq.py
+++ b/api/handlers/iq.py
@@ -3,6 +3,7 @@ import base64
 import io
 import logging
 from typing import List, Optional
+from multiprocessing import Pool
 
 from blob.azure_client import AzureBlobClient
 from database import datasource_repo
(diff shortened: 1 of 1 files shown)


api/handlers/iq.py
```diff
@@ -68,33 +68,84 @@ async def get_iq_data(
...
+            block_index_chunk,
+            block_size,
+            bytes_per_iq_sample,
+            iq_file,
+            azure_client,
+            blob_size,
         )
+        for block_index_chunk in block_indexes_chunks
+    ]
+    return await asyncio.gather(*tasks)
```
Nepomuceno, line 119: Since this does not use pools wouldn't this be actually sending potentially 800 requests at the same time to the storage account? Would this get us easily to be throttle because Of too many concurrent requests?

Jade-Codes: We could use semaphore to limit requests:
https://stackoverflow.com/questions/48483348/how-to-limit-concurrency-with-python-asyncio

Nepomuceno: I think that could work defintly!

Jade-Codes: added semaphore to limit requests

api/handlers/iq.py
```diff
@@ -68,33 +68,91 @@ async def get_iq_data(
...
+
+    if blob_size < offsetBytes:
+        return b""
+    if blob_size < offsetBytes + countBytes:
+        countBytes = blob_size - offsetBytes
+    content = await azure_client.get_blob_content(
+        filepath=iq_file, offset=offsetBytes, length=countBytes
+    )
+
+    st = time.time()
```
Nepomuceno, line 152: WOuld make sense to move this to the begining of the function.

Nepomuceno (approved):
image

PR #427, opened 2023-08-04
Access control for data sources

PaulDFoster:
### What does this PR do?
Access control implementation for data sources.
Using a JWT provided by the Authorization header, two routines provide access control for data sources.
data source documents in the mongodb are extended with owners and readers lists containing preferred usernames (emails) or AAD groups, and a bool flag Public indication public access to the data source. Owners indicates administrative control of the data source such as updating, deleting and setting other user access levels.
get_current_user - validates the JWT and returns a dictionary of its contents, including roles, groups and user name.
check_access - takes data source id (account + container) and evaluates the get_current_user dict groups against those defined in data source documents.

In API methods where access to a single data source is being attempted check_access can be made a dependency, and its result evaluated appropriately within the API method.
In API methods where access to multiple data sources is being attempted get_current_user can be made a dependency and its result can be appropriately evaluated within the API method using calls to check_access for each data source access attempt. This is required to enable result sets to contain more than one data source result.

If no Authorization header (including JWT) is provided access will be determined based each data source's Public bool flag. If a data source is marked as Public it will be accessible.
New data sources are created with the calling user's preferred username if a Authorization header is provided, and/or the IQEngine-Admin group as an Owner. This ensures data sources are read-only when marked as public access but administration tasks can then be performed.

No data source administrative UI is provided in this PR.
Only datasource apis protected in this PR. Others will be implemented as separate PR

The 3 Item "None" of "Optional[Any]" has no attribute "get_secret_value" linter errors are being ignored as a None object cannot reach this point of execution.

### All Submissions:

### Changes to Core Features:


api/handlers/iq.py
```diff
@@ -236,8 +238,10 @@ async def get_iq_data_slices(
     filepath: str,
     datasource: DataSource = Depends(datasource_repo.get),
     azure_client: AzureBlobClient = Depends(AzureBlobClient),
-    current_user: Optional[dict] = Depends(required_roles()),
+    access_allowed=Depends(check_access)
 ):
+    if access_allowed is None:
```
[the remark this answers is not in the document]
Jade-Codes: Or should be an explicit 'unauthorized' returned

client/src/api/datasource/api-client.ts
```diff
@@ -2,20 +2,66 @@ import axios from 'axios';
...
+        account: this.account,
+        scopes: [api_scope],
+      });
+      return response.accessToken;
+    } catch (error) {
+      return null;
+    }
+  }
+
+  private async requestWithAuth(config) {
```
Jade-Codes, line 29: Not sure this accurately describes the intention, maybe something like 'requestWithRequiredHeaders' or similar

client/src/api/datasource/api-client.ts
```diff
@@ -2,20 +2,66 @@ import axios from 'axios';
...
+    return axios({ ...config, headers: {} });
+  }
+
   async sync(account: string, container: string): Promise<void> {
-    await axios.put(`/api/datasources/${account}/${container}/sync`);
+    await this.requestWithAuth({
+      method: 'get',
+      url: `/api/datasources/${account}/${container}/sync`,
+    });
+    //await axios.put(`/api/datasources/${account}/${container}/sync`);
```
Jade-Codes, line 45: pls remove these commented out methods

client/src/api/datasource/datasource-client-factory.ts
```diff
@@ -4,15 +4,23 @@ import { BlobClient } from './blob-client';
...
 export const DataSourceClientFactory = (
   type: string,
   files: FileWithDirectoryAndFileHandle[],
   dataSources: Record<string, DataSource>
 ): DataSourceClient => {
   switch (type) {
-    case CLIENT_TYPE_API:
-      return new ApiClient();
+    case CLIENT_TYPE_API: {
+      const { instance } = useMsal();
```
Jade-Codes, line 16: why not just move this to the ApiClient?

client/src/api/datasource/api-client.ts
PaulDFoster: Disagree. The headers are generally provided. The procedure specifically adds the authorization header making a call with auth.

api/handlers/iq.py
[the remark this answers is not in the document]
PaulDFoster: It returns the level of permission. So a string for a defined level or None for none.

client/src/api/datasource/api-client.ts
Jade-Codes: But it doesn't always request with auth?

PaulDFoster: If no bearer is provided in config it does not

Jade-Codes: That's what I mean by the name is confusing

Jade-Codes (approved):

PR #447, opened 2023-08-08
Specified "array" in plugins API types

SamBr0wn:


plugins/plugins_openapi.yaml
```diff
@@ -78,7 +78,7 @@ paths:
         - in: path
           name: functionName
           schema:
-            type: string
+            type: string9
```
Nepomuceno, line 81: Probably not intentional

Nepomuceno (commented):
Doing this changes will require to change the current implementaiton that we do have of plugins to accpet that and change the frontend to also send the data in this format.

plugins/plugins_openapi.yaml
SamBr0wn: Yes, this makes sense. This was purely a semantics thing, but if it will impact the frontend then it's not worth it. And yes typo there.

777arc:
currently prod is using the same plugins server as staging so making an API change will have to be done sparingly and in bursts, so I'll make sure this change gets included the next time we mess with plugin API.

PR #505, opened 2023-08-11
removed or commented out several logs

777arc:
Since a lot of work is still in progress, I prefer to comment these out instead of removing them altogether, at least for the time being.


Nepomuceno (approved):
We need to remember to remove this after

PR #506, opened 2023-08-11
Move api docs endpoint to api_docs

777arc:
so it doesnt conflict with our actual docs, when we run api and client on the same port


Nepomuceno (approved):
LGTM

PR #517, opened 2023-08-14
Ignore the models dir when finding available plugins

777arc:
image

### DIFF
--- a/plugins/src/plugins_api.py
+++ b/plugins/src/plugins_api.py
@@ -36,6 +36,7 @@ async def get_plugins_list():
             dirs.append(file)
     dirs.remove("__pycache__")
     dirs.remove("template_plugin")
+    dirs.remove("models")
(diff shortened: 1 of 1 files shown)


plugins/src/plugins_api.py
```diff
@@ -36,6 +36,7 @@ async def get_plugins_list():
             dirs.append(file)
     dirs.remove("__pycache__")
```
Nepomuceno, line 37: I would change this code for 

for file in os.listdir("."):
        d = os.path.join(".", file)
        if os.path.isdir(d) and os.path.isfile(os.path.join(d, file + ".py")):
            dirs.append(file)
    if "__pycache__" in dirs:
        dirs.remove("__pycache__")
    if "template_plugin" in dirs:
        dirs.remove("template_plugin")
    return dirs

PR #530, opened 2023-08-15
Adjusted workflow permissions to address security concerns

777arc:


.github/workflows/run_staging_deployment.yml
```diff
@@ -14,17 +14,17 @@ jobs:
   create_pre_release:
     uses: ./.github/workflows/docker-pre-release.yml
     permissions:
-      contents: write
+      contents: read
       pull-requests: write
-      packages: write
+      packages: read
```
Jade-Codes, line 19: doesn't this need to be write?

777arc: nice catch, fixed

Nepomuceno (approved):
LGTM

PR #532, opened 2023-08-15
time cursor, toggle entire sample

jh87692:
### What does this PR do?
allows the selection of the entire sample (via time cursor bounds) to push entirety to plugin server

### All Submissions:

### Changes to Core Features:


Nepomuceno:
That would not work now because the cursor would try to download the whole file to the browser immediately

Nepomuceno:
That would not work now because the cursor would try to download the whole file to the browser immediately

777arc:
Yeah I think a new checkbox in the plugin section called "entire recording" and disabled by default would be nice.  I was also meaning to make it so when you don't have the time cursors enabled, it just uses whatever span is visible on your screen, up to you if you want to also add that.

jh87692:
> Yeah I think a new checkbox in the plugin section called "entire recording" and disabled by default would be nice. I was also meaning to make it so when you don't have the time cursors enabled, it just uses whatever span is visible on your screen, up to you if you want to also add that.

yeah that makes sense. I'll modify it tomorrow

PR #548, opened 2023-08-16
Remove connectionInfo and allow cloud storage on public repos

hannahlspencer:
### What does this PR do?
- Removes connectionInfo and replaces it with data source information from the API
- Adds functionality for public data sources to be used with cloud storage

This code will change with the introduction of the storage key being used in the backend.

### All Submissions:

### DIFF
--- a/client/src/pages/recording-view/components/plugins-pane.tsx
+++ b/client/src/pages/recording-view/components/plugins-pane.tsx
@@ -19,7 +19,6 @@ import { dataTypeToBytesPerIQSample } from '@/utils/selector';
 import { getDataSource } from '@/api/datasource/queries';
 import { useSpectrogramContext } from '../hooks/use-spectrogram-context';
 import { useCursorContext } from '../hooks/use-cursor-context';
-import { CLIENT_TYPE_API } from '@/api/Models';
 
 export enum MimeTypes {
   ci8 = 'iq/ci8',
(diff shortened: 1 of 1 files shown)


client/src/pages/recording-view/components/plugins-pane.tsx
```diff
@@ -44,10 +45,15 @@ export const PluginsPane = () => {
   const [modalSamples, setModalSamples] = useState<Float32Array>(new Float32Array([]));
   const [modalSpectrogram, setmodalSpectrogram] = useState(null);
   const [useCloudStorage, setUseCloudStorage] = useState(true);
-  const { dataSourcesQuery } = useUserSettings();
-  const connectionInfo = dataSourcesQuery?.data[`${account}/${container}`];
+  const dataSource = getDataSource(CLIENT_TYPE_API, account, container, true).data;
```
Nepomuceno, line 48: We need to be careful this page might still be used with a blob datasource or with a local file so pining up this here might not be a good idea

Jade-Codes, line 48: I think:

const {data }  = getDataSource(CLIENT_TYPE_API, account, container, true);

Is usually the preferred way of retrieving a property :)

client/src/pages/recording-view/components/plugins-pane.tsx
```diff
@@ -305,7 +335,7 @@ export const PluginsPane = () => {
           ))}
         </select>
       </label>
-      {connectionInfo && (
+      {dataSource && (
```
Nepomuceno, line 338: We should probably do a different check here for exemole you don't want that for local files but you do want this for personal blob storage

Nepomuceno (commented):
We need to make sure it still works for local files and for local blob storage.

client/src/pages/recording-view/components/plugins-pane.tsx
```diff
@@ -96,15 +102,39 @@ export const PluginsPane = () => {
...
+          },
+        ],
+        custom_params: {
+          start_freq: annotation ? annotation['core:freq_lower_edge'] : null,
+          end_freq: annotation ? annotation['core:freq_upper_edge'] : null,
+        },
+      };
+    }
+
+    if (useCloudStorage && dataSource) {
```
Jade-Codes, line 129: I would just do:

if (useCloudStorage) 

once.

change sas_token to be null for now, then when the implementation for 'get_blob_token' is done, tie that in

client/src/pages/recording-view/components/plugins-pane.tsx
hannahlspencer: To get the sas token (for now)

Jade-Codes: I would remove as we really shouldn't be using any form of sas token from the data sources or connection info going forward... Not sure why we are returning this from the API

PR #553, opened 2023-08-17
temp sas generation on api server

jh87692:
### What does this PR do?
generates a temp sas token of 1 hour duration if account key exists in the data source. If it doesn't and the sas token is present, it will return this (if authorized). For public sources, the sas token will be returned - do we want this?

### All Submissions:

### Changes to Core Features:


api/handlers/datasources.py
```diff
@@ -159,3 +161,39 @@ async def sync_datasource(
...
+    existing_datasource = await datasources_collection.find_one(
+        {
+            "account": account,
+            "container": container,
+        }
+    )
+    if not existing_datasource:
+        raise HTTPException(status_code=404, detail="Datasource not found")
+    if not existing_datasource.get("account_key", None):
+        if not existing_datasource["sasToken"] or access_allowed == "public":
```
hannahlspencer, line 185: Looking at it now, I think we should split up the access allowed + no sas token check, just in case there's no sas token and there should be. There could be a case where public access isn't allowed but for some reason the datasource wasn't given an sas token on upload, so we should differentiate between those when telling the front end why there isn't a token so the user can know if there's an error. Maybe if public access allowed we can just return a blank token or null or something which should be fine, and then an error if public access isn't allowed and there's no sas

jh87692: ok, I've added the separation

hannahlspencer (approved):

api/handlers/datasources.py
hannahlspencer: cheers!

PR #555, opened 2023-08-17
DRAFT: Enable large zoom levels

tobybrad:
### What does this PR do?

Builds on the work done on the data fetching methods to allow larger portions of the file to be viewed than was previously possible.

The principal here is that zoom levels now represent what portion of the file the user is asking to view. We calculate the fftSkip value required to meet that need. The starting point is always 0, don't skip and currently we increment the zoom levels by 10% each step until 100% i.e. view the whole recording.

This PR is currently draft for the following reasons:

- Are linear 10% increments useful? Not much visual difference between 50% and 100% tbh. Maybe a curve?
- If you drag the window down to the bottom of the recording and set an extreme zoom we try to display past the end of the file. Should we move the window back up the recording or some other strategy (prevent the zoom?!?!)
- There's a small gap at the bottom of the image when setting zoom to 100%. Need to look into that (or set the yellow background to black in PROD builds so you don't see it?)

### All Submissions:

### Changes to Core Features:


Nepomuceno (approved):
LGTM!

Nepomuceno:
1) Are linear 10% increments useful? Not much visual difference between 50% and 100% tbh. Maybe a curve?
Last chat I had with @777arc the suggested ranges for the fft was [1, 2, 5, 10, 20, 50, 100, 1000, 10k, 100k]

2) If you drag the window down to the bottom of the recording and set an extreme zoom we try to display past the end of the file. Should we move the window back up the recording or some other strategy (prevent the zoom?!?!)

I think we should cap the max level to the zoom to something that display 100% of the record in the screen.

3) the colour it si due to incomplete FFTS meaning that somethig. that has a partial fft in the end of the file.

PR #557, opened 2023-08-18
DRAFT Minimap cleanup and size is now static (300px) to speed up initial page load time

777arc:
It's a balance between quality of the minimap and how long it takes the page to load when you go to a new recording, and it was taking quite a while when it required 800-1000 FFTs, now with 300 FFTs it feels way snappier and the minimap still looks pretty good imo.

Also fixed bug where there was 1 extra row at the bottom of zeros

Also changed requiredFFTs to minimapFFTIndices

### DIFF
--- a/client/src/api/iqdata/Queries.ts
+++ b/client/src/api/iqdata/Queries.ts
@@ -71,15 +71,15 @@ export function useGetIQData(
 
   const queryClient = useQueryClient();
   const { filesQuery, dataSourcesQuery } = useUserSettings();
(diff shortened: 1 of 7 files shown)


Nepomuceno:
I think it would be more beneficial for us to actually implement the creation of the minimap in the backend that way we could save a lot on the number of requests that we do to the blob storage and avoid accidentaly hitting the blob storage limits.

I will see what I can get going on that but could be an alternative.

PR #612, opened 2023-08-31
allow for multiple non-IQ output files from plugin

sebastus:
### What does this PR do?
The new audio detection / speech-to-text plugin returns multiple wav files but only one of them gets downloaded. This PR changes the behavior such that all wav files are downloaded.

### All Submissions:

### Changes to Core Features:

### DIFF
--- a/client/src/pages/recording-view/components/plugins-pane.tsx
+++ b/client/src/pages/recording-view/components/plugins-pane.tsx
@@ -142,6 +142,15 @@ export const PluginsPane = () => {
       }
     }
 
+    const BlobFromSamples = (samples_base64, data_type) => {
+      const samples = window.atob(samples_base64);
+      var blob_array = new Uint8Array(samples.length);
+      for (var i = 0; i < samples.length; i++) {
+        blob_array[i] = samples.charCodeAt(i);
+      }
+      return new Blob([blob_array], { type: data_type });
+    }
+
     fetch(selectedPlugin, {
       method: 'POST',
       headers: {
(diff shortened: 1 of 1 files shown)


client/src/pages/recording-view/components/plugins-pane.tsx
```diff
@@ -224,31 +233,38 @@ export const PluginsPane = () => {
...
-            const samples = window.atob(samples_base64);
-            var blob_array = new Uint8Array(samples.length);
-            for (var i = 0; i < samples.length; i++) {
-              blob_array[i] = samples.charCodeAt(i);
-            }
-            const a = document.createElement('a');
-            const blob = new Blob([blob_array], { type: data.data_output[0]['data_type'] });
-            const url = window.URL.createObjectURL(blob);
+            let d = new Date();
+            let datestring = d.getFullYear() + "-" + ("0" + (d.getMonth() + 1)).slice(-2) + "-" + ("0" + d.getDate()).slice(-2)
```
Nepomuceno, line 237: Isnt better to just use .toISOString()

sebastus: It comes pretty close, but you get 2011-10-05T14:48:00.000Z. I didn't want the decimal seconds or Z at the end. And I didn't want to research what it generates in all circumstances. So I could have used iso string and split at the '.' and took [0], but I didn't know if that would always work. 

Would it?

sebastus: looks like (new Date().toISOString()).split('.')[0] will do the trick.

PR #633, opened 2023-09-04
fix data loaded indicator

sebastus:
### What does this PR do?

fixes microscopic bug with data loaded indicator

array.find returns the found index, not a boolean. if the found index is zero, it's falsy, so the existing code is broken in that case.

### All Submissions:

### Changes to Core Features:

### DIFF
--- a/client/src/pages/recording-view/components/scroll-bar.tsx
+++ b/client/src/pages/recording-view/components/scroll-bar.tsx
@@ -75,8 +75,8 @@ const ScrollBar = ({ currentFFT, setCurrentFFT }: ScrollBarProps) => {
     const tiles = [];
     const downloadScaling = meta.getTotalSamples() / fftSize / 100;
     for (let i = 0; i < 100; i++) {
-      const exist = downloadedIndexes.find((x) => x >= i * downloadScaling && x < (i + 1) * downloadScaling);
-      if (exist) {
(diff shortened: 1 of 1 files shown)


client/src/pages/recording-view/components/scroll-bar.tsx
```diff
@@ -75,8 +75,8 @@ const ScrollBar = ({ currentFFT, setCurrentFFT }: ScrollBarProps) => {
     const tiles = [];
     const downloadScaling = meta.getTotalSamples() / fftSize / 100;
     for (let i = 0; i < 100; i++) {
-      const exist = downloadedIndexes.find((x) => x >= i * downloadScaling && x < (i + 1) * downloadScaling);
-      if (exist) {
+      const foundIndex = downloadedIndexes.find((x) => x >= i * downloadScaling && x < (i + 1) * downloadScaling);
+      if (foundIndex != undefined) {
```
Jade-Codes, line 79: should this me (foundIndex && foundIndex != undefined)? Just wondering what would happen if it was 'null'

sebastus: I don't think array.find() can return a null, but doesn't hurt anything. fixed.

Jade-Codes (approved):

PR #749, opened 2023-11-21
implemented .wav converter

romanziske:
Implemented a first simple .wav to SigMF converter. Currently, the left and right channels are embedded in a complex float 32, so the SigMF file has only one channel.


777arc:
I added an integration test that goes through all the steps of uploading/downloading, it verifies the name of the downloaded file but unfortunately I couldn't figure out how to get it to verify the contents of the zip, which would be nice because if the backend conversion fails it seems to provide an empty/invalid zip file.

777arc:
Nice!  Anything else you wanted to do before merging this guy?

777arc:
Awesome it should show up on staging.iqengine.org in ~5 mins

PR #767, opened 2023-12-02
Moved wav-to-sigmf converter to separate file and sped it up

777arc:
@romanziske mind reviewing this?  since it lays the framework for other converters.  I actually found it super useful to call directly from python when I had to convert some 4GB local files from wav to SigMF earlier today.


romanziske (approved):
Looks good. It should be easier to add more converters now. 

When we have multiple converters, we could also move the converters to their own repo and build a proper cli application, as upload/download will always be a bottleneck for larger files.

777arc:
Yeah and can be published to pypi, then used as a dep by IQEngine

777arc:
But we shouldnt rush to do that because it's still runnable from CLI, just not as streamlined as it could be.

PR #796, opened 2024-01-17
reworked plugin api spec for async processing

romanziske:
This PR contains a **first draft** for the async plugin openapi spec. 

Notable changes:
- /{functionName} now works with multipart
- /{functionName} returns a 'jobID' (eg. a uuid) to get the results in async manner
- /{jobID}/result get the result of a given job (similar to old post request to run plugin)
- /{jobID}/status get the status/progess of a given job

Polling is required if the plugin in ran from an frontend client like IQEngine, if for any reason a server starts a plugin run, it would be nice to pass a callback url to the plugin that will be called when the plugin run is done.


777arc:
for /{jobID}/result should there be a 3rd response, for when the job isnt done yet, or maybe it could redirect to the status endpoint?

777arc:
Even though this is a draft I'm going to merge it in, since this `plugins_openapi.yaml` has essentially always been a draft, and it will continue to be until I update the python server to match it which isn't going to happen that soon anyway, there is another party interested in getting involved in this API design and I want them to see the current version on main branch

romanziske:
> for /{jobID}/result should there be a 3rd response, for when the job isnt done yet, or maybe it could redirect to the status endpoint?
I agree with that. Including a third response could help eliminate the need for the /status endpoint, as you mentioned earlier. I'm unsure about the appropriate response code or if it's considered good practice, though. When I think about it, instead of redirecting, it might be better to have it return the status directly.

romanziske:
> for /{jobID}/result should there be a 3rd response, for when the job isnt done yet, or maybe it could redirect to the status endpoint?

I agree with that. Including a third response could help eliminate the need for the /status endpoint, as you mentioned earlier. I'm unsure about the appropriate response code or if it's considered good practice, though. When I think about it, instead of redirecting, it might be better to have it return the status directly.

PR #801, opened 2024-01-24
Minor improvements to support Airbus SigMF Files

romanziske:
This pr introduces 2 changes to properly display Airbus SigMF file:

- only draw visible annotation => sped up drawing for dataset with many annotations
- support to draw annotation relative to capture center frequency


777arc:
Nice!!!

777arc (approved):

PR #844, opened 2024-03-09
Async Plugins Server and bump pydantic+fastapi

romanziske:
This is a draft PR for the async plugin API. 
It already implements async processing, progress tracking, and the result endpoint. 
Another significant change is the refactoring of the plugin dataclass, which is now an abstract class that handles progress tracking, error handling, and result storage. This way, a new Plugin only needs to implement the rf_function itself.

Some open points:
- [x] support custom params
- [ ] support for object storage
- [x] better handling of results (e.g., don't return IQ result as JSON) => its not possible to return file + json, so for now keep binary data as base64 encoded string in result
- [x] integration with the frontend


romanziske:
made some progress on the plugins. also added example to simply mount custom plugins in to plugin api docker container.  any chance we can make the plugin stuff as an pypi and npm package to use it outside of the repo?

777arc:
> made some progress on the plugins. also added example to simply mount custom plugins in to plugin api docker container. any chance we can make the plugin stuff as an pypi and npm package to use it outside of the repo? @777arc

Mmm good idea, which one would you say is better to tackle first?  I can add the GitHub action

romanziske:
> > made some progress on the plugins. also added example to simply mount custom plugins in to plugin api docker container. any chance we can make the plugin stuff as an pypi and npm package to use it outside of the repo? @777arc
> 
> Mmm good idea, which one would you say is better to tackle first? I can add the GitHub action

i think the python plugin stuff is more important, so people can build plugins based on the new plugin class. The have the react stuff would be nice to easily integrate the plugin api into IQDMs viewer, but for now i can just copy the code. 

But with the python plugin available, we can serve our ml stuff as a plugin. The docker image can be for the plugin server can be hosted on github like the current base image?

777arc:
Sorry for taking so long to dive into this!!

777arc:
I love the new code structure!  Don't worry about the straight-from-cloud functionality, better we get this PR finished and merged in, then I can add it back in.

777arc:
Between GNU Radio and SatDump and future large deps we add to the plugins base-image, at some point I'll add a way to choose which deps are included in the build, like with flags, but until then are you OK with just leaving it as 1 base-image (that includes all) in order to reduce complexity?  Or did you need the smaller gnuradio-free image for your own stuff?

777arc (approved):

777arc:
I'm OK merging as long as we make an Issue or something to remember what is left to add

romanziske:
> Between GNU Radio and SatDump and future large deps we add to the plugins base-image, at some point I'll add a way to choose which deps are included in the build, like with flags, but until then are you OK with just leaving it as 1 base-image (that includes all) in order to reduce complexity? Or did you need the smaller gnuradio-free image for your own stuff?

Yes Lets just keep one base image, i only Needed it for testing.

> I'm OK merging as long as we make an Issue or something to remember what is left to add

Ok i will create an issue

PR #858, opened 2024-04-04
Fix authentication of query page

Jade-Codes:
When trying to get the information for the IQData and Metadata, an error was being thrown because the token was returning a bad request.

This was because an 'undefined' scope was being passed to it.

I have updated the code to pass the correct configuration in.


777arc:
Waiting to merge till I verify the other modes work, like using files local to backend

Jade-Codes:
> Waiting to merge till I verify the other modes work, like using files local to backend

Okay no worries, I can't imagine it affecting anything because all it does it retrieve the token if the user is logged in. 

But famous last words 😃

PR #866, opened 2024-04-16
Added download button when viewing a recording

javAlborz:


777arc:
You could try using the existing downlink links that are on the browser page, eg https://github.com/IQEngine/IQEngine/blob/main/client/src/pages/browser/file-row.tsx#L131

most of the meat is here https://github.com/IQEngine/IQEngine/blob/main/client/src/utils/sigmfMetadata.ts#L107

although I see you're also trying to zip it before the download, which might not be worth the complexity because zipping binary IQ files usually doesnt lead to much savings in disk space.

777arc:
btw feel free to ping me on discord if you want help with any of this, eg testing it or whatnot

777arc:
oh actually I think this is one of those bots that tries to solve github issues with ML, clearly failing in this case

javAlborz:
Hi Marc. This is my first github issue, but would be happy yo try to help

-Alborz

javAlborz:
Hi Marc. Sent you a message on Discord would be happy yo try to help out

777arc:
> Hi Marc. Sent you a message on Discord would be happy yo try to help out

Not seeing anything on Discord

PR #885, opened 2024-05-29
Async to Sync plugin proxy

loic-fejoz:
While IQEngine is transitioning from Sync plugins to Async one, this pull-request propose a plugin that bridge the gap for existing plugin. It has been tested with my current sync plugin in rust as of 2024-05-29.

This plugin can proxy only function at a time. The url of the proxied sync plugin is configured by the environment variable `IQENGINE_PROXY_URL`.

For instance, if your `userdef` sync function plugin is served on port 9000, one can launch the following command:
`IQENGINE_PROXY_URL=http://127.0.0.1:9000/plugins/userdef uvicorn --host 0.0.0.0 --port 8000 plugins_ap`

### DIFF
--- a/plugins/src/plugins/sync_proxy/sync_proxy.py
+++ b/plugins/src/plugins/sync_proxy/sync_proxy.py
@@ -9,6 +9,7 @@ import os
 import urllib.request
 import requests
 
+display_url = os.environ.get('IQENGINE_PROXY_DISPLAY', 'False').lower() in ['true', '1', 't', 'y', 'yes']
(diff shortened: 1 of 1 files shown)


777arc:
FYI I just deployed the latest main to prod (iqengine.org) so now both staging and prod are on the async

777arc:
This was a great solution!

loic-fejoz:
Thanks @777arc . I thus need to hurry up to transition to my plugin too!

PR #920, opened 2024-08-26
Bugfix: Fix diffSeconds calculation for time cursor

jeremy-epiq:
- Divide cursorTime difference (end - start) by sample rate


777arc:
ah nice catch, ty!!!

jeremy-epiq:
Happy to help!

PR #938, opened 2024-09-09
Implements CIF1-3

SimonWittkowski:


777arc:
Based on the very first change I read, `This converter converts Vita49 compliant data into SigMF. Change the input path of the Vita49 file to the location of your` it seems like this wasn't reviewed, so closing for now

PR #1006, opened 2025-10-07
Update converter.tsx to fix the link to the converter README.md in main instead of "speed-up-wav-converter".

KelseyCreekSoftware:
This PR is to fix the link to the converter README.md, by pointing to main, instead of "speed-up-wav-converter". Not sure on the history here, so there may be a better readme to link to, but the one in main makes sense to me.


777arc:
Nice catch, ty!

